PR #10, opened 2020-06-13
Small tweaks

brmscheiner:


locale/en.json
```diff
@@ -394,9 +394,8 @@
     "SERVER_TOOLTIP_JOBCOUNTER_LABEL": "Jobcounter:",
     "SERVER_TOOLTIP_STATUS_LABEL": "Status:",
     "SERVER_TOOLTIP_DATE_RECEIVED_LABEL": "Received on:",
-    "SERVER_TOOLTIP_ARIA_LABEL": "{jobCounterLabel} {jobCounterValue}; {statusLabel} {statusValue}; {dateReceivedLabel} {dateReceivedValue}",
```
brmscheiner, line 397: There's not really anything to translate here

src/pages/individual/PictureBook.jsx
```diff
@@ -82,20 +82,9 @@ export default function PictureBook() {
...
-      if (
-        isLoading &&
-        !Object.values(pagesLoading).some(value => value === true)
-      ) {
-        setIsLoading(false);
-      }
-    },
-    [isLoading, pagesLoading],
+  const isLoading = Object.values(pagesLoading).some(value =>
+    Boolean(value),
```
brmscheiner, line 86: Actually `.some(value => value)` would work the same too, but it's less clear

PR #18, opened 2020-07-26
transitions

Emily-Ke:
- The layout on the report sightings page is not pixel perfect to what it was before.
- I had to change the way the photo credit on the 404 page displayed because, after changing the layout for the transitions, it displayed on top of the footer. The way it is right now, it should always display over the picture, but not always within the initial screen view.


brmscheiner:
looks great

PR #88, opened 2021-02-12
Build the package nightly, on push and conditional on pull request

mmulich:
Build the package in as a continuous activity of development.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,44 @@
+name: Build
+on:
+  schedule:
+    - cron: '0 0 * * *' # Every day at midnight
+  push:
+  pull_request:
+    paths:
+      - '.github/workflows/build.yml'
+      # Build when a fundamental package change has occured
+      - 'package.json'
+
+jobs:
+  build:
+    name: Build the frontend releasable package
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Set up Node.js
+        uses: actions/setup-node@v2.1.4
+        with:
+          node-version: '15.x'
+
+      - name: Install dependencies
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -0,0 +1,44 @@
...
+      - name: Set up Node.js
+        uses: actions/setup-node@v2.1.4
+        with:
+          node-version: '15.x'
+
+      - name: Install dependencies
+        run: npm install
+
+      - name: Copy template api keys file
+        run: cp src/constants/apiKeysTemplate.js src/constants/apiKeys.js
```
brmscheiner, line 29: The frontend will be really buggy without real API keys. Is it possible to generate a legit apiKeys.js using GitHub secrets?

brmscheiner (approved):
Great to get this ball rolling but it should be noted that the frontend will just show a "server unreachable" error page if it's not able to connect to a Houston backend. We may also want some more refinement around apiKeys.js

.github/workflows/build.yml
mmulich: I'm not sure that's wise. I was hoping to discuss this with you at a later time.

It seems to me that the frontend should default to getting these API keys from houston. In a development environment without a backend, maybe it cascades to checking for a specific file location.

My hope would be that a compiled build can be deployed without any changes. Hence the reason for getting these secret'ish bits from houston. This would also allow all configuration to be done on the backend without doing a redeployment of the frontend, which is always a nice-to-have.

Those are just some thoughts... We can talk about them more later.

brmscheiner: Hmm, yes, I think when things are fleshed out administrators would enter their API keys as site settings. The one exception to this rule might be the translation service, I need to give that some thought soon..

PR #123, opened 2021-07-12
Add src/constants/apiKeysTemplate.js back

karenc:
`apiKeysTemplate.js` was removed in commit
"DEX-337 create AlertWrapper component" (86226532) and it was probably a
mistake (not explained in commit message).  Add the file back because it
is necessary for the build.


brmscheiner (approved):
Thankk you!!

PR #124, opened 2021-07-14
DEX-374 add and remove some links in the footer and resize the width …

Atticus29:
…of menu list at certain screen pixel lengths. Flexbox is difficult to divide evenly: https://stackoverflow.com/questions/22085646/can-flexbox-divide-items-evenly-on-multiple-rows

I hate this solution.
But it does look like using media queries or the multi-column module (which looked like a confusing refactor given ListMenu's dependence on flexbox): https://stackoverflow.com/questions/22085646/can-flexbox-divide-items-evenly-on-multiple-rows.

Other people have solved this problem by adding n-1 "hidden" flexbox items. I played around with that a little but ultimately abandoned it because 
{schema.map(item => (
        <MenuItem key={item.id}>
          <Link
            href={get(data, [item.customHref, 'value']) || item.href}
            external={item.external}
          >
            <Text id={item.id} />
          </Link>
        </MenuItem>
      ))}
      
a) wouldn't let me insert more than one html element within the body of the map function and 
b) I couldn't figure out how to easily include only n-1 rather than n "hidden" MenuItems and
c) it seemed _even less elegant_ than the other solutions.


Atticus29:
@brmscheiner I also have some questions about strategies for arriving at appropriate pix values for .medium-resize elements and for the @media screen and (min-width: 703px) and (max-width: 900px) restrictions. Remind me to ask you. Might be easier on share screen.

src/styles/globalStyles.css
```diff
@@ -9,3 +9,9 @@
 .MuiTimelineItem-missingOppositeContent:before {
   flex: unset !important;
 }
+
+@media screen and (min-width: 703px) and (max-width: 900px) {
```
brmscheiner, line 13: Adjusting the width of the element with a media query seems like the way to go! But I do try to minimize using the global stylesheet - material UI provides a utility for doing this kind of thing within the component:

https://material-ui.com/components/use-media-query/

brmscheiner (approved):

PR #136, opened 2021-08-12
Dex 345 autocomplete regions

Atticus29:


src/components/fields/edit/LocationIdEditor.jsx
```diff
@@ -8,8 +10,20 @@ import TreeItem from '@material-ui/lab/TreeItem';
 import useLabel from '../../../hooks/useLabel';
 import useDescription from '../../../hooks/useDescription';
 import FormCore from './FormCore';
+import TextField from '@material-ui/core/TextField';
+import { FormattedMessage } from 'react-intl';
```
brmscheiner, line 14: react-intl is already imported on line 2

src/components/fields/edit/LocationIdEditor.jsx
```diff
@@ -8,8 +10,20 @@ import TreeItem from '@material-ui/lab/TreeItem';
...
+import Text from '../../Text';
 
-export default function TreeViewInput(props) {
+const useStyles = makeStyles((theme) => ({
+  root: {
+    width: 500,
+    '& > * + *': {
```
brmscheiner, line 20: very fancy lol

src/components/fields/edit/LocationIdEditor.jsx
```diff
@@ -20,67 +34,46 @@ export default function TreeViewInput(props) {
...
-      onChange(nodeIds);
-    } else {
-      onChange(nodeIds.filter(id => id !== 'IGNORE'));
-    }
-    setSelected(nodeIds);
-  };
+  function collapseChoices(choices, depth) {
+    const result = choices.map(choice => {
+      if (!choice.locationID) {
+        return { depth, name: ' '.repeat(depth) + choice.name + (getNumDescendents(choice) ? '(' + getNumDescendents(choice).toString() + ')' : ''), id: choice.id };
```
brmscheiner, line 50: this is too much for one line i think:
a) don't forget to run the formatter before pushing (`npm run format`)
b) defining some variables above the return would make it more legible i think
c) since we are doing the spacing in the css (on line 64) we could eliminate the spaces right?
d) dont forget about those template literals! they are much better than building strings this way 

```
const numberOfDescendants = getNumDescendants(choice);
const descendantsString = numberOfDescendants ? ` (${numberOfDescendants})` : '';
const name = choice.name + descendantsString;
return { depth, name, id: choice.id};
```

package.json
```diff
@@ -4,7 +4,7 @@
   "description": "",
   "main": "index.js",
   "scripts": {
-    "start": "run-script-os",
+    "start": "run-script-os --no-cache",
```
brmscheiner, line 7: what's this for?

Atticus29: cruft from an attempt to fix. My bad.

brmscheiner (approved):

PR #140, opened 2021-08-18
DEX-390 add roles to the edit user table and make filterable

Atticus29:


Atticus29:
Rendering the roles into a column isn't readily reusable.
1. Would we want to code for the possibility of wanting that to be reusable?
2. How might you advise doing something like that in React?

brmscheiner (commented):
So if you wanted to make the logic you have reusable, it would be a good idea to start a folder of table cell renderers in the common components area. That way you could just import `roleCellRenderer` (for instance) and do `customBodyRenderer: roleCellRenderer`

However I think in this case there is more happening than there should be. Why compute the labels from the roles every time, when each role could just have a known label? In fact they already exist in `roleSchema.js`

So you can do
```
  customBodyRender: roles => {
    const roleLabels = roles.map(r => {
      const matchingRole = roleSchema.find(schema => schema.id === r);
      return matchingRole.titleId;
    })
    return roleLabels.join(', ');
  }
```

brmscheiner (approved):

PR #158, opened 2021-09-16
Dex 390 fix lowercase query

Atticus29:
The only real meaningful change I made was:
if (
        dataValue &&
        (dataValue.includes(filter) ||
          dataValue.includes(filter.toLowerCase()))
      )
  
in DataDisplay.jsx


src/components/dataDisplays/DataDisplay.jsx
```diff
@@ -104,7 +104,12 @@ export default function DataDisplay({
               .trim()
           : JSON.stringify(rawValue);
       }
-      if (dataValue && dataValue.includes(filter)) match = true;
+      if (
+        dataValue &&
+        (dataValue.includes(filter) ||
+          dataValue.includes(filter.toLowerCase()))
+      )
```
brmscheiner, line 111: Since dataValue is lowercased above, `dataValue.includes(filter.toLowerCase()))` must be true whenever `dataValue.includes(filter)` is true. So you can simplify:

```
if (
        dataValue && dataValue.includes(filter.toLowerCase()))
      )
```

brmscheiner (approved):

PR #180, opened 2021-11-03
DEX-404 change button, div, and sightingCard data display styling in …

Atticus29:
…response to QA feedback

Note: The sighting card looks good (IMO) in browser view, but I don't love how it looks in mobile view, but I'm not sure what is best practice on that front. Open to advice, if this is a big enough fish to fry currently.

### DIFF
--- a/src/components/Button.jsx
+++ b/src/components/Button.jsx
@@ -74,6 +74,7 @@ function Core({
 
   if (display === 'link') {
     roleStyles = {
+      textTransform: 'none',
       background: 'none!important',
       border: 'none',
       padding: '0!important',
@@ -81,7 +82,24 @@ function Core({
       textDecoration: 'underline',
       cursor: 'pointer',
       backgroundColor: 'transparent',
+      padding: 0,
+      textAlign: 'right',
+      justifyContent: 'right',
+      fontSize: '14px',
     };
+    return (
+      <button
+        type="button"
+        style={{ ...roleStyles, ...style }}
+        {...rest}
+      >
+        {loading ? (
+          <CircularProgress size={24} style={{ color: 'white' }} />
+        ) : (
+          children
+        )}
+      </button>
+    );
   }
 
   if (disabled) {
--- a/src/components/Button.jsx
+++ b/src/components/Button.jsx
@@ -75,14 +75,13 @@ function Core({
   if (display === 'link') {
     roleStyles = {
       textTransform: 'none',
-      background: 'none!important',
+      background: 'none',
       border: 'none',
-      padding: '0!important',
+      padding: '0',
       color: theme.palette.text.primary,
       textDecoration: 'underline',
       cursor: 'pointer',
       backgroundColor: 'transparent',
-      padding: 0,
       textAlign: 'right',
       justifyContent: 'right',
       fontSize: '14px',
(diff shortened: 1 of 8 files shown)


src/components/cards/SightingsCard.jsx
```diff
@@ -179,7 +179,7 @@ export default function SightingsCard({
   );
 
   return (
-    <div>
+    <div style={{ marginTop: '0.5em', marginLeft: '0.5em' }}>
```
brmscheiner, line 182: so we actually dont use `em`s in our project, we use a 4px grid system. since `em` is the font size, 0.5em will usually work out to 7px which doesn't match the grid

src/components/cards/SightingsCard.jsx
```diff
@@ -214,6 +214,7 @@ export default function SightingsCard({
       >
         {encounters && (
           <DataDisplay
+            style={{ width: '37em' }}
```
brmscheiner, line 217: same issue here

src/components/Button.jsx
```diff
@@ -74,14 +74,32 @@ function Core({
...
+      fontSize: '14px',
     };
+    return (
+      <button
+        type="button"
+        style={{ ...roleStyles, ...style }}
+        {...rest}
+      >
+        {loading ? (
+          <CircularProgress size={24} style={{ color: 'white' }} />
```
brmscheiner, line 97: i know it's silly with white and black but let's use `theme.palette.common.white` here. trying to keep off-theme colors out of the codebase.

src/components/Button.jsx
```diff
@@ -74,14 +74,32 @@ function Core({
 
   if (display === 'link') {
     roleStyles = {
+      textTransform: 'none',
       background: 'none!important',
       border: 'none',
       padding: '0!important',
```
brmscheiner, line 80: i'm fairly certain the importants are no longer necessary since we're not overriding anything anymore

brmscheiner (commented):
Took a look at the card width problem.. looks like the reason the default card styling was messing up was because of the wrapper div around everything. So I was able to eliminate all of the new styles along with that div and get a good result:
```
  return (
    <Card
      title={title}
      titleId={titleId}
      renderActions={
        <div>
          <IconButton
            style={{ color: theme.palette.primary.main }}
            aria-label="View list"
          >
            <ViewList />
          </IconButton>
          <IconButton aria-label="View chart">
            <ViewMap />
          </IconButton>
        </div>
      }
    >
      <StandardDialog
        open={modalOpen}
        onClose={onClose}
        titleId="GPS_TITLE"
      >
        <DialogContent style={{ marginBottom: 24 }}>
          <MapInSighting
            latitude={get(gpsCoordinates, 'lat')}
            longitude={get(gpsCoordinates, 'lng')}
          />
        </DialogContent>
        <DialogActions style={{ padding: '0px 24px 24px 24px' }}>
          <Button id="CLOSE" display="basic" onClick={onClose} />
        </DialogActions>
      </StandardDialog>
      {encounters && (
        <DataDisplay
          style={{}}
          noTitleBar
          columns={filteredColumns}
          data={encountersWithLocationData}
        />
      )}
    </Card>
  );
```
image

src/components/cards/SightingsCard.jsx
Atticus29: @brmscheiner Would you recommend doing my margins in pixels instead? Would that be wonky, depending on screen size? 

Perhaps this will be moot after the above gets resolved.

src/components/Button.jsx
Atticus29: I took the liberty of doing this everywhere.

brmscheiner (approved):
lgtm!

PR #182, opened 2021-11-04
Dex 486 collaboration management

Atticus29:


brmscheiner (commented):
image
I noticed that if a user does not have a full_name, they still show up in the dropdown but as a small bar with no text (I have one highlighted in the screenshot above). Let's do Unnamed User, with the email address in parentheses if possible.

brmscheiner:
I also noticed that after creating a collaboration it looks weird for collaborating user 1...

User 1 sees this:
image


User 2 sees this:
image

User 1 should be allowed to view user 2's data at this point, and the collaboration looks good, so I'm wondering if just the table needs to be fixed? Feel free to create a new ticket and/or PR for this if you would rather...

brmscheiner:
Let's move the dropdowns next to each other with a small amount of margin between them, instead of having them spread apart from each other 
image

src/pages/collaborations/collaborationManagementForm.jsx
```diff
@@ -0,0 +1,126 @@
...
+          style={{ marginBottom: '20px' }}
+          loading={buttonLoading}
+          onClick={async () => {
+            setButtonLoading(true);
+            const successful = await establishCollaboration(
+              //need the await here. Otherwise, setShouldDisplay(true) below fires asynchronously
+              user1,
+              user2,
+            );
+            setButtonLoading(false);
```
brmscheiner, line 95: this works, but the better way is to get the loading status from the `useEstablishConnection` hook. so instead of keeping track of the state here you can do 

```
  const {
    establishCollaboration,
    loading,
    error,
    success,
  } = useEstablishCollaborationAsUserManager();
```

and then 
```
<Button loading={loading} ....... />
```

src/pages/userManagement/UserManagement.jsx
```diff
@@ -22,6 +22,7 @@ import usePostUser from '../../models/users/usePostUser';
 import useGetUsers from '../../models/users/useGetUsers';
 import UserEditTable from './UserEditTable';
 import roleSchema from './constants/roleSchema';
+import CollaborationManagementForm from '../collaborations/collaborationManagementForm'; // TODO not sure whether this is where the form component should live. Likely /components instead??
```
brmscheiner, line 25: it's ok where you have it, since it is only used on one page. The `/components` directory is mainly for UI that is repeated in multiple places

brmscheiner (approved):

PR #183, opened 2021-11-15
Improve sightings table

brmscheiner:
- remove extra border around map 
- change table size to match collaborations table 
- include link to sighting in table 

image
image

### DIFF
--- a/src/components/cards/SightingsCard.jsx
+++ b/src/components/cards/SightingsCard.jsx
@@ -21,7 +21,7 @@ import Card from './Card';
 export default function SightingsCard({
   title,
   titleId = 'SIGHTINGS',
-  encounters,
+  sightings,
   columns = ['date', 'location', 'actions'],
   onDelete,
 }) {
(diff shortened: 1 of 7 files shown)


src/components/cards/SightingsCard.jsx
```diff
@@ -83,9 +83,9 @@ export default function SightingsCard({
     return '';
   };
 
-  const encountersWithLocationData = useMemo(
+  const sightingsWithLocationData = useMemo(
     () => {
-      const encsWithFormattedLocations = encounters.map(
+      const encsWithFormattedLocations = sightings.map(
```
Atticus29, line 88: while we're in the process of naming variables better, makes sense to do this one, too?

src/components/cards/SightingsCard.jsx
```diff
@@ -83,9 +83,9 @@ export default function SightingsCard({
...
         currentEncounter => ({
```
Atticus29, line 89: same here

Atticus29 (approved):
LGTM... My only recommendations are cosmetic.

PR #184, opened 2021-11-15
DEX-505 lift state and use uppy hook to prevent creation of multiple …

Atticus29:
…uppy instances and lose track of assets upon click of back button

1. I'm not sure whether currentExifData stuff also needs it's state lifted out of UploadManager (how can I test whether the exif stuff is working correctly?).
2. See "TODO" comment in the code for other question.


Atticus29:
Two notable things happened here:
1. I had to lift state of the uppy instance to the parent component.
2. I had to use the useUppy hook (h/t to a college friend for these tips).

PR #190, opened 2021-11-19
Patch AGS properties

brmscheiner:
This PR makes it possible to patch assetGroupSighting top-level metadata fields. It also does a few more things I will note in the comments

### DIFF
--- a/src/components/cards/SightingsCard.jsx
+++ b/src/components/cards/SightingsCard.jsx
@@ -157,7 +157,7 @@ export default function SightingsCard({
           <div>
             <ActionIcon
               variant="view"
-              href={`/individuals/${value}`}
+              href={`/sightings/${value}`}
             />
             {onDelete && (
               <ActionIcon
--- a/src/components/fields/view/SelectViewer.jsx
+++ b/src/components/fields/view/SelectViewer.jsx
@@ -23,7 +23,7 @@ export default function SelectViewer({
     intl.formatMessage({ id: 'OPTION_LABEL_NOT_FOUND' });
   let label = get(selectedOption, labelKey, translatedDefaultLabel);
   if (value === null || value === undefined) {
-    label = intl.formatMessage({ id: 'Value not set' });
+    label = intl.formatMessage({ id: 'VALUE_NOT_SET' });
   }
 
   return (
--- a/src/models/sighting/useSightingFieldSchemas.js
+++ b/src/models/sighting/useSightingFieldSchemas.js
@@ -111,12 +111,12 @@ export default function useSightingFieldSchemas() {
           descriptionId: 'SIGHTING_END_TIME_DESCRIPTION',
           category: defaultSightingCategories.general.name,
         }),
-        createFieldSchema(fieldTypes.string, {
-          name: 'verbatimEventDate',
-          labelId: 'SIGHTING_VERBATIM_TIME',
-          descriptionId: 'SIGHTING_VERBATIM_TIME_DESCRIPTION',
-          category: defaultSightingCategories.general.name,
-        }),
+        // createFieldSchema(fieldTypes.string, {
+        //   name: 'verbatimEventDate',
+        //   labelId: 'SIGHTING_VERBATIM_TIME',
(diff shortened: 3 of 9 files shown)


src/components/cards/SightingsCard.jsx
```diff
@@ -157,7 +157,7 @@ export default function SightingsCard({
           <div>
             <ActionIcon
               variant="view"
-              href={`/individuals/${value}`}
+              href={`/sightings/${value}`}
```
brmscheiner, line 160: this fixes a bug in the sightings table that was causing all the links to 404

src/components/fields/view/SelectViewer.jsx
```diff
@@ -23,7 +23,7 @@ export default function SelectViewer({
     intl.formatMessage({ id: 'OPTION_LABEL_NOT_FOUND' });
   let label = get(selectedOption, labelKey, translatedDefaultLabel);
   if (value === null || value === undefined) {
-    label = intl.formatMessage({ id: 'Value not set' });
+    label = intl.formatMessage({ id: 'VALUE_NOT_SET' });
```
brmscheiner, line 26: this was a basic translation error I made in an earlier PR

src/models/sighting/useSightingFieldSchemas.js
```diff
@@ -111,12 +111,12 @@ export default function useSightingFieldSchemas() {
           descriptionId: 'SIGHTING_END_TIME_DESCRIPTION',
           category: defaultSightingCategories.general.name,
         }),
-        createFieldSchema(fieldTypes.string, {
-          name: 'verbatimEventDate',
-          labelId: 'SIGHTING_VERBATIM_TIME',
-          descriptionId: 'SIGHTING_VERBATIM_TIME_DESCRIPTION',
-          category: defaultSightingCategories.general.name,
-        }),
+        // createFieldSchema(fieldTypes.string, {
```
brmscheiner, line 114: temporarily disabled the `verbatimEventDate` field until DEX-562 is complete. otherwise sighting PATCH is broken

src/pages/sighting/EditSightingMetadata.jsx
```diff
@@ -28,14 +29,27 @@ export default function EditSightingMetadata({
...
   } = usePatchSighting();
 
+  const {
+    updateProperties: updateAgsProperties,
+    loading: agsLoading,
+    error: agsError,
+    setError: setAgsError,
+  } = usePatchAssetGroupSighting();
+
+  const error = pending ? agsError : sightingError;
```
Atticus29, line 48: It's not clear to me how pending modulates which of these gets assigned to error.

Atticus29: Same for the three lines below this.

src/pages/sighting/StatusCard.jsx
```diff
@@ -22,6 +22,8 @@ export default function StatusCard({ sightingData }) {
   const dateString = dateCreated
     ? formatDate(dateCreated, true)
     : 'unknown date';
+  const detectionComplete = get(sightingData, 'stage') !== 'detection';
```
Atticus29, line 25: Is there any 'stage' that could be an error? Is it necssarily true (and will it be forever) that !=='detection' will me that detection is complete?

Atticus29 (approved):
Not sure whether the questions I have are approval-preventing or not, so pre-approving this for you to decide whether any of this requires more work on your part.

src/pages/sighting/EditSightingMetadata.jsx
brmscheiner: oh, if an AGS is being viewed, then `updateAgsProperties` will always be used instead of `updateSightingProperties`. and so on

src/pages/sighting/StatusCard.jsx
brmscheiner: i dont actually know what all the stages even are! this statusCard is going to be fleshed out much more soon i think. for now we dont have the data to do much with it

PR #192, opened 2021-11-29
New bao

brmscheiner:


src/ErrorBoundary.jsx
```diff
@@ -37,96 +38,90 @@ class ErrorBoundary extends React.Component {
...
+            </Accordion>
+          </div>
         </div>
       </div>
     );
   }
 }
 
-export default ErrorBoundary;
+export default withTheme(ErrorBoundary);
```
Atticus29, line 127: Had to look this up. Cool/weird.

src/components/cards/SightingsCard.jsx
```diff
@@ -84,14 +84,19 @@ export default function SightingsCard({
   };
 
   const sightingsWithLocationData = useMemo(
-    () =>
-      sightings.map(sighting => ({
+    () => {
+      // hotfix //
```
Atticus29, line 88: Not sure why this comment is necessary.

src/components/cards/SightingsCard.jsx
```diff
@@ -84,14 +84,19 @@ export default function SightingsCard({
...
+      if (!sightings) return [];
+      // hotfix //
+
+      return sightings.map(sighting => ({
```
Atticus29, line 92: So _this_ is how you'd do that. Noted.

src/components/dataDisplays/DataDisplay.jsx
```diff
@@ -117,6 +125,8 @@ export default function DataDisplay({
     visibleColumnNames.includes(column.name),
   );
 
+  const noResults = showNoResultsBao && data && data.length === 0;
```
Atticus29, line 128: Is there any possibility for data to be falsey here? If so, maybe something like (howNoResultsBao && data && data.length === 0) || !data
 ?

src/pages/auth/Forgot.jsx
```diff
@@ -48,7 +48,7 @@ export default function Forgot() {
     <SimpleFormPage
```
Atticus29, line 48: Why this SimpleFormPage component twice?

src/pages/auth/RequestInvitation.jsx
```diff
@@ -44,7 +45,8 @@ export default function RequestInvitation() {
     <SimpleFormPage
```
Atticus29, line 45: Same question here.

src/pages/individual/Individual.jsx
```diff
@@ -12,6 +12,11 @@ import useIndividual from '../../models/individual/useIndividual';
 import useDeleteIndividual from '../../models/individual/useDeleteIndividual';
 import usePatchIndividual from '../../models/individual/usePatchIndividual';
 
+// VERILY BAD HOTFIX //
+import defaultIndividualSrc from '../../assets/defaultIndividual.png';
```
Atticus29, line 16: ???

src/pages/individual/EditIndividualMetadata.jsx
```diff
@@ -37,6 +37,10 @@ export default function EditIndividualMetadata({
     setError,
   } = usePatchIndividual();
 
+  // hotfix //
```
Atticus29, line 40: I don't understand why this is a hotfix if the product isn't hot yet.

src/pages/individual/Individual.jsx
Atticus29: 1. What's so bad about this?
2. Why is this a hotfix?

src/ErrorBoundary.jsx
brmscheiner: yeah these class components are the "old way" react used to work. however for some reason it appears error boundaries are still supposed to use them? hence the difference for this one component

src/components/dataDisplays/DataDisplay.jsx
brmscheiner: yes, but the `&& data` is the part where that is accounted for. cool thing about javascript is when you have multiple `&&` operators strung together, it will short circuit and return false for the whole expression without evaluating the rest of the line. so if `data=undefined` here you actually never get the `undefined.length does not exist` error

src/pages/auth/Forgot.jsx
brmscheiner: There are sort of two different pages in this component! 
1. forgot password? give us your email
2. email sent!

src/pages/individual/Individual.jsx
brmscheiner: hmm, really these should be "temporary fixes". i made some quick adjustments to render the individuals page (which currently is erroring out) so that i could get the bao working on that page. but i need to go back and do it the _right_ way once i move on to the individuals tickets 

these particular one is really bad because it's using components from the wrong page

PR #193, opened 2021-11-30
Address QA issues with report sighting form

brmscheiner:


src/components/fields/edit/DateEditor.jsx
```diff
@@ -40,43 +33,24 @@ export default function DateEditor(props) {
...
         <div style={{ display: 'flex', flexDirection: 'column' }}>
-          <KeyboardDatePicker
+          <KeyboardDateTimePicker
             disableToolbar
             disableFuture
             variant="inline"
             autoOk
-            format="yyyy-MM-dd" // US: MM/dd/yyyy
+            ampm={false} // US: true
+            format="yyyy-MM-dd HH:mm" // US: MM/dd/yyyy hh:mm
```
Atticus29, line 42: What modulates whether this get's the U.S. treatment vs. not? Here, it looks hard-coded to _not_ be U.S. format.

src/components/fields/edit/LatLongEditor.jsx
```diff
@@ -24,14 +35,33 @@ export default function LatLongEditor({
...
+    currentLongitudeString,
+    setCurrentLongitudeString,
+  ] = useState(initialLongitudeString);
+
+  useEffect(
+    () => {
+      const {
+        latitudeString,
+        longitudeString,
+      } = deriveGpsStringsFromValue(value);
```
Atticus29, line 56: I don't understand why the useMemo above is necessary if this derive method gets run every time the values change anyway.

src/components/fields/edit/LatLongEditor.jsx
```diff
@@ -46,7 +76,13 @@ export default function LatLongEditor({
           value={currentLatitudeString}
           onChange={e => {
             const inputValue = e.target.value;
-            onChange([parseFloat(inputValue), currentLongitude]);
+            const floatValue = parseFloat(inputValue);
+            if (Number.isNaN(floatValue)) {
+              onChange([parseFloat(inputValue), currentLongitude]);
+            } else {
+              onChange([null, currentLongitude]);
```
Atticus29, line 83: Looks like this line and the line below it could possibly be a mix up between latitude and longitude?

src/components/fields/edit/LatLongEditor.jsx
```diff
@@ -56,7 +92,13 @@ export default function LatLongEditor({
           value={currentLongitudeString}
           onChange={e => {
             const inputValue = e.target.value;
-            onChange([currentLatitude, parseFloat(inputValue)]);
+            const floatValue = parseFloat(inputValue);
+            if (Number.isNaN(floatValue)) {
+              onChange([currentLatitude, parseFloat(inputValue)]);
+            } else {
+              onChange([currentLatitude, null]);
```
Atticus29, line 99: Maybe not.... Seems unlikely that you would make this mistake twice. Possibly disregard the above.

src/components/fields/edit/DateEditor.jsx
brmscheiner: for MVP all times and dates are hardcoded to be international. but later this will be a user preference

src/components/fields/edit/LatLongEditor.jsx
brmscheiner: well the useMemo prevents it from happening twice. since this useMemo has no dependencies it will only get run once on the component's first render

src/components/fields/edit/LatLongEditor.jsx
brmscheiner: it's correct as-is actually! since the latitude is changing, onChange is recording the new latitude value plus the current (unchanged) longitude value

PR #194, opened 2021-11-30
Remove redux from project

brmscheiner:
This was where almost all the fake data was stored. For the most part we have real APIs where I was using redux selectors, and for the rest I just put empty data for now.


src/components/report/TermsAndConditionsDialog.jsx
```diff
@@ -1,15 +1,13 @@
...
-import { selectTermsAndConditions } from '../../modules/site/selectors';
 import Text from '../Text';
 import StandardDialog from '../StandardDialog';
 
 export default function TermsAndConditionsDialog({
   visible,
   onClose,
 }) {
-  const termsAndConditions = useSelector(selectTermsAndConditions);
+  const termsAndConditions = 'These are our terms: liberty or death!';
```
Atticus29, line 10: Lol... I imagine that this does not need to go into en.json?

src/index.jsx
```diff
@@ -21,15 +19,9 @@ if (!__DEV__) {
...
-const { store } = storeConfigs;
-
 const root = document.getElementById('root');
 
-const Main = hot(() => (
-  <Provider store={store}>
-    <App />
-  </Provider>
-));
+const Main = hot(() => <App />);
```
Atticus29, line 24: So <Provider>'s only purpose is to provide the redux store?

src/pages/bulkImport/TermsAndConditionsDialog.jsx
```diff
@@ -1,12 +1,11 @@
...
 import DialogContent from '@material-ui/core/DialogContent';
-import { selectTermsAndConditions } from '../../modules/site/selectors';
 import Text from '../../components/Text';
 import StandardDialog from '../../components/StandardDialog';
 
-export default function EditAvatar({ visible, onClose }) {
-  const termsAndConditions = useSelector(selectTermsAndConditions);
+export default function TermsAndConditionsDialog({ visible, onClose })
+{
+  const termsAndConditions = 'Here are our terms!!!';
```
Atticus29, line 8: Not sure if this should connect to en.json.

src/pages/org/Orgs.jsx
```diff
@@ -1,28 +1,19 @@
...
 
 export default function User() {
   const intl = useIntl();
-  const siteName = useSelector(selectSiteName);
-  const pageTitle = intl.formatMessage(
-    { id: 'ORGS_PAGE_TITLE' },
-    { siteName },
-  );
-  useDocumentTitle(pageTitle, { translateMessage: false });
+  useDocumentTitle('ORGANIZATIONS');
```
Atticus29, line 13: I'm not sure whether it's a problem that this is not assigned to a variable.

src/pages/org/Orgs.jsx
```diff
@@ -1,28 +1,19 @@
...
 
-  const orgs = useSelector(selectOrgs);
-  const isAdministrator = useSelector(selectIsAdministrator);
+  const orgs = [];
+  const isAdministrator = true;
```
Atticus29, line 16: This might become easy to lose track of. Maybe we can add a TODO comment here or connect this to a config file or something? 

In other words, I presume that we won't always want this to be true forever, but I'm  not sure what we should do to make sure this doesn't become a bug.

src/pages/projects/Projects.jsx
```diff
@@ -1,23 +1,20 @@
...
-import { selectIsAdministrator } from '../../modules/app/selectors';
 import useDocumentTitle from '../../hooks/useDocumentTitle';
 import CreateProject from './CreateProject';
 import fakeProjects from './fakeProjects';
 
 export default function Projects() {
   const intl = useIntl();
-  const pageTitle = intl.formatMessage({ id: 'PROJECTS_PAGE_TITLE' });
-  useDocumentTitle(pageTitle, { translateMessage: false });
+  useDocumentTitle('PROJECTS');
```
Atticus29, line 15: Again, not sure whether this should be assigned to a variable.

src/pages/projects/Projects.jsx
```diff
@@ -1,23 +1,20 @@
...
 
-  const isAdministrator = useSelector(selectIsAdministrator);
+  const isAdministrator = true;
```
Atticus29, line 17: See earlier comment.

src/pages/user/Users.jsx
```diff
@@ -12,17 +11,12 @@ import Text from '../../components/Text';
...
 export default function Users() {
   const intl = useIntl();
-  const siteName = useSelector(selectSiteName);
-  const pageTitle = intl.formatMessage(
-    { id: 'USERS_PAGE_TITLE' },
-    { siteName },
-  );
-  useDocumentTitle(pageTitle, { translateMessage: false });
+
+  useDocumentTitle('USERS');
```
Atticus29, line 19: See earlier comment

Atticus29 (approved):
Don't think any of my comments are deal-breakers, but take a look just in case.

src/components/report/TermsAndConditionsDialog.jsx
brmscheiner: No.. I pinged Tanya about actually creating these and then we can incorporate them correctly

src/index.jsx
brmscheiner: The store provider, yes

src/pages/org/Orgs.jsx
brmscheiner: Thing is the orgs UI isn't even part of MVP. So this whole file is more or less TODO status. Along with projects

PR #196, opened 2021-12-03
Dex 536 sightings card map mode

Atticus29:


brmscheiner (commented):
image
only one pin should be open at a time (opening a new pin should close previous pin)
image
remove shadow that appears when hovering over a pin

image
text should not be all caps, container should be wider

image
map icon should glow when map is selected (and table icon should not glow)

image
i dont think we need the word "location" or the container in the container here. just the map

src/components/cards/SightingMapViewCard.jsx
```diff
@@ -0,0 +1,37 @@
...
+import Card from './Card';
+
+export default function SightingMapViewCard({
+  title,
+  titleId = 'LOCATION',
+  data,
+}) {
+  const intl = useIntl();
+  const transformedData = data
+    .filter(entry => get(entry, 'decimalLatitude', null) !== null)
```
brmscheiner, line 16: you can do `.filter(entry => get(entry, 'decimalLatitude')` for the same result

src/components/cards/SightingMapViewCard.jsx
```diff
@@ -0,0 +1,37 @@
+import { get } from 'lodash';
+import React from 'react';
+import { useIntl, FormattedMessage } from 'react-intl';
+
+import { formatDate } from '../../utils/formatters';
+import ManyPoints from '../maps/ManyPoints';
+import Card from './Card';
+
+export default function SightingMapViewCard({
```
brmscheiner, line 9: since this is already rendered inside a card, we don't need a card here. i think `SightingMap` would be fine

src/components/cards/SightingsCard.jsx
```diff
@@ -202,23 +212,35 @@ export default function SightingsCard({
...
             <ViewList />
           </IconButton>
-          <IconButton aria-label="View chart">
+          <IconButton
+            disabled={
+              sightingsWithLocationData.filter(entry =>
+                get(entry, 'decimalLatitude'),
+              ).length < 1
+            }
+            aria-label="View chart"
```
brmscheiner, line 225: view map

src/components/maps/ManyPoints.jsx
```diff
@@ -0,0 +1,77 @@
...
+            if (currentLat > northernmostLat || !northernmostLat)
+              northernmostLat = currentLat;
+            if (currentLong < westernmostLong || !westernmostLong)
+              westernmostLong = currentLong;
+            if (currentLong > easternmostLong || !easternmostLong)
+              easternmostLong = currentLong;
+          }
+        });
+        const sw = new maps.LatLng(southernmostLat, westernmostLong);
+        const ne = new maps.LatLng(northernmostLat, easternmostLong);
```
brmscheiner, line 51: we might wind up moving this calculation to a utility function for use in other maps

brmscheiner: (but its fine here for now)

src/components/maps/ManyPoints.jsx
```diff
@@ -0,0 +1,77 @@
...
+        map.setZoom(Math.max(map.getZoom() - 1, 0)); // add some buffer space to it
+      }}
+    >
+      {latLongLabelArr.map(entry => {
+        return (
+          <Marker
+            entry={entry}
+            clickAwayClicked={clickAwayClicked}
+            setClickAwayClicked={setClickAwayClicked}
+            style={{ outlineColor: 'red' }}
```
brmscheiner, line 66: these aren't showing up red for me

brmscheiner (approved):

PR #197, opened 2021-12-03
Ags encounter metadata

brmscheiner:


src/models/encounter/useEncounterFieldSchemas.js
```diff
@@ -61,6 +57,11 @@ export default function useSightingFieldSchemas() {
         };
       });
 
+      speciesOptions.push({
+        label: intl.formatMessage({ id: 'UNKNOWN' }),
```
Atticus29, line 61: Not clear to me why this should be hard-coded as unknown.

src/models/encounter/usePatchEncounter.js
```diff
@@ -0,0 +1,84 @@
...
+      );
+      dictionaryCopy.decimalLongitude = get(
+        dictionaryCopy,
+        ['gps', 1],
+        null,
+      );
+      delete dictionaryCopy.gps;
+    }
+
+    // delete dictionaryCopy.taxonomy;
```
Atticus29, line 27: Why commented out? Just a bookmark for the future?

Atticus29 (commented):
Generally, all looks good!

I'm curious why region isn't available in the encounter-level metadata edit:
Screen Shot 2021-12-03 at 2 32 47 PM
y

src/models/encounter/usePatchAgsEncounter.js
```diff
@@ -0,0 +1,86 @@
...
+  const [error, setError] = useState(null);
+  const [success, setSuccess] = useState(false);
+
+  const updateProperties = async (
+    assetGroupSightingId,
+    encounterId,
+    dictionary,
+  ) => {
+    const dictionaryCopy = { ...dictionary };
+    if ('gps' in dictionaryCopy) {
```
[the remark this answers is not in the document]
brmscheiner: yes!

src/models/encounter/useEncounterFieldSchemas.js
brmscheiner: it's just an option in the species dropdown to remove the existing species. so for example if they accidentally put zebra they can switch it back to unknown to remove the species property

src/models/encounter/usePatchEncounter.js
brmscheiner: i should remove this actually..

PR #199, opened 2021-12-06
More Bao

brmscheiner:
Switch login screen to solid bao 

Success Bao on report screens
Screen Shot 2021-12-06 at 1 42 59 PM
Screen Shot 2021-12-06 at 1 42 44 PM


Atticus29:
The context here was indeed helpful!

Atticus29 (approved):
All looks good to me. A thought that just occurred to me: was it a mistake to call these things Bao in the code base? What if we move away for Bao-looking mascots in the future?

Would we have been better off calling them something more generic like mascot or avatar?

At least Bao is a somewhat search-and-replaceable word, I suppose.

PR #202, opened 2021-12-08
Location ID fixes 

brmscheiner:
Not sure what bugs are being caused by the old code but it seems wrong to me


src/components/maps/Marker.jsx
```diff
@@ -16,9 +15,7 @@ export default function Marker({
   useEffect(
     () => {
       const thisMarkersSightingId = get(entry, 'guid');
-      currentMarkerToShow === thisMarkersSightingId
-        ? setShowInfo(true)
-        : setShowInfo(false);
+      setShowInfo(currentMarkerToShow === thisMarkersSightingId);
```
Atticus29, line 18: Nice. Thanks!

Atticus29 (approved):

Atticus29:
LGTM. Tested locally and looks like it behaves well.

PR #203, opened 2021-12-09
QA fixes

brmscheiner:


src/components/fields/edit/LatLongEditor.jsx
```diff
@@ -10,14 +10,18 @@ import Button from '../../Button';
 import StandardDialog from '../../StandardDialog';
 import useDescription from '../../../hooks/useDescription';
 
+function getNumberString(n) {
+  if (n === 0) return '0';
```
brmscheiner, line 14: the bugs caused by typing 0 were occurring because 0 is falsey. so just short-circuiting that problem here.

Atticus29 (approved):
LGTM

PR #204, opened 2021-12-09
Integrate react-query

brmscheiner:


src/App.jsx
```diff
@@ -37,34 +40,7 @@ const ScrollToTop = function() {
   return null;
 };
 
-function reducer(state, action) {
```
brmscheiner, line 40: we will no longer need react context hooks with this update

src/components/RequestCollaborationButton.jsx
```diff
@@ -42,8 +46,7 @@ export default function RequestCollaborationButton({ otherUserId }) {
   return (
     <Button
       onClick={async () => {
-        const successful = await requestCollaboration(otherUserId);
-        if (successful) refresh();
```
brmscheiner, line 46: `refresh` is no longer called explicitly. instead, we will invalidate the stale query when objects are updated (many examples in code)

brmscheiner:
Main purpose of this PR is to optimize load times.

*Before*
https://user-images.githubusercontent.com/9775178/145499404-565d9d55-dc84-42db-94e4-65bd0b42965f.mov

*After*
https://user-images.githubusercontent.com/9775178/145499423-6bddc65a-3ce9-43f9-9c59-9f0efbffbc49.mov

Secondary purpose is to make fetch code cleaner. 

This isn't a full integration, but these two endpoints were our worst offenders from an optimization standpoint. We will observe whether react-query causes problems for a week or so then move forward with the migration. In the meantime, any new fetches should be done with react-query

src/models/collaboration/useRequestCollaboration.js
```diff
@@ -25,6 +29,7 @@ export default function useRequestCollaboration() {
         setSuccess(true);
         setError(null);
         setLoading(false);
+        queryClient.invalidateQueries(queryKeys.me);
```
Atticus29, line 32: I don't understand why queryKeys.me specifically was the one that had to be invalidated here.

src/models/site/usePutSiteSettings.js
```diff
@@ -22,7 +24,7 @@ export default function usePutSiteSettings() {
       const successful = get(response, ['data', 'success'], false);
       setLoading(false);
       if (successful) {
-        dispatch(setSiteSettingsNeedsFetch(true));
+        queryClient.invalidateQueries(queryKeys.settingsConfig);
```
Atticus29, line 27: So, after invalidation, how do these query keys become valid again?

src/models/site/useSiteSettings.js
```diff
@@ -1,100 +1,63 @@
...
-  );
-  const [schemaLoading, setSchemaLoading] = useState(
-    siteSettingsSchemaNeedsFetch,
+export default function useSiteSettings() {
+  const settingsSchemaResult = useQuery(
+    queryKeys.settingsSchema,
+    async () => {
+      const response = await axios({
+        url: `${__houston_url__}/api/v1/configurationDefinition/default/__bundle_setup`,
+        timeout: 2000,
```
Atticus29, line 13: Why the timeout?

src/models/site/useSiteSettings.js
```diff
@@ -1,100 +1,63 @@
...
+      });
+      return get(response, 'data.response.configuration');
+    },
+    {
+      staleTime: Infinity,
```
Atticus29, line 18: ?

src/models/users/useGetMe.js
```diff
@@ -1,53 +1,36 @@
...
-
-      getMe();
+  const result = useQuery(
+    queryKeys.me,
+    async () => {
+      const response = await axios.request({
+        url: `${__houston_url__}/api/v1/users/me`,
+        method: 'get',
+      });
+      if (!__DEV__) {
```
Atticus29, line 15: This seems like a python thing?

src/models/users/useGetMe.js
```diff
@@ -1,53 +1,36 @@
...
+        Sentry.setUser({
```
Atticus29, line 16: Why is Sentry.setUser necessary? What happens if this is ommitted?

src/models/users/usePatchUser.js
```diff
@@ -1,9 +1,13 @@
 import { useState } from 'react';
 import axios from 'axios';
 import { get } from 'lodash-es';
+import { useQueryClient } from 'react-query';
+
+import queryKeys from '../../constants/queryKeys';
 import { formatError } from '../../utils/formatters';
 
 export default function usePatchUser(userId) {
+  const queryClient = useQueryClient();
```
Atticus29, line 10: I'm seeing that some of these custom hooks are using a QueryClient while others are not. I'm sure that this will become more clear to me once I read the documentation, but I find myself wondering why.

src/pages/settings/Settings.jsx
```diff
@@ -29,7 +29,7 @@ function getInitialFormValues(schemas, data) {
 export default function Settings() {
   useDocumentTitle('SETTINGS_AND_PRIVACY');
 
-  const { data, refresh } = useGetMe();
+  const { data } = useGetMe();
```
Atticus29, line 32: No invalidate statements in this file?

Atticus29: Same with a few of the subsequent modified files.

Atticus29 (approved):
Just a few things I'm confused about. I doubt any of them will impact the PR.

src/models/site/usePutSiteSettings.js
brmscheiner: "invalidating" the settings config is basically saying "the settings config is now stale because we just updated it"

telling react-query this will cause it to automatically begin refetching the settings config

src/models/collaboration/useRequestCollaboration.js
brmscheiner: after creating a collaboration we refresh `api/v1/me` because that's where the collaborations are stored

src/models/site/useSiteSettings.js
brmscheiner: this is copied over from the old request. at some point i must have deemed this necessary because the request was taking a long time to resolve?

src/models/site/useSiteSettings.js
brmscheiner: you'll have to take a look at the `react-query` docs for this one

src/models/users/useGetMe.js
brmscheiner: `__DEV__` is an environment variable created by webpack. it's just a boolean that is true if the frontend is compiled for development false if its compiled for production

src/models/users/useGetMe.js
brmscheiner: Sentry is our error reporting. this just makes it so we can see the username of whoever is encountering errors, which will hopefully help with support

brmscheiner: by the way this is just coming from the old code on the left side, not new to this PR

src/models/users/usePatchUser.js
brmscheiner: it's just like when we do `const intl = useIntl();`

src/pages/settings/Settings.jsx
brmscheiner: no, so the invalidate statements go in the hooks that update resources, and as a result the `refresh` commands are no longer needed in the components themselves

PR #209, opened 2021-12-12
Removing unwanted black fills

541joehill:
Explicitly added `fill="none"` to paths for this component that were filling with black.


brmscheiner (approved):
looks great, thanks!

PR #211, opened 2021-12-14
Show initial value in latlng input

brmscheiner:
Fixes a bug where LatLng input would not show its initial value


Atticus29 (approved):
LGTM

PR #212, opened 2021-12-15
Fix region editor

brmscheiner:
Fixes a couple of bugs:
 - Selecting a subregion was throwing an error on the report form page (`cannot read property "id" of undefined`)
 - Dropdown options were changing after selecting a region 
 
Added an optimization so that choices are not recursively collapsed every render

### DIFF
--- a/src/components/fields/edit/LocationIdEditor.jsx
+++ b/src/components/fields/edit/LocationIdEditor.jsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useMemo } from 'react';
 import FormHelperText from '@material-ui/core/FormHelperText';
 import { Autocomplete } from '@material-ui/lab';
 import { TextField } from '@material-ui/core';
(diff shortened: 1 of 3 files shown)


src/components/fields/edit/LocationIdEditor.jsx
```diff
@@ -36,46 +36,51 @@ export default function LocationIdEditor(props) {
...
+    : null;
+  const currentRegion = get(currentRegionArray, [0], null);
 
   return (
     <FormCore schema={schema} width={width}>
       <Autocomplete
-        value={
-          value
-            ? get(
-                get(schema, 'choices').filter(
```
brmscheiner, line 71: the biggest issue was here: the collapsed choices needed to be filtered, wasn't working on the nested object

src/components/fields/edit/LocationIdEditor.jsx
```diff
@@ -94,9 +99,9 @@ export default function LocationIdEditor(props) {
           }
         }}
         getOptionLabel={option => get(option, 'name', '')}
-        getOptionSelected={(option, val) => {
-          return option.id ? option.id === val.id : false;
-        }}
+        getOptionSelected={option =>
```
Atticus29, line 102: Why make this independent of val? I don't understand what benefit is gained by doing it this way.

Atticus29 (approved):
LGMT. Just comments/questions.

src/components/fields/edit/LocationIdEditor.jsx
brmscheiner: Same difference, I think. I just changed it as I was debugging and never changed it back since it was working well

PR #219, opened 2021-12-31
DEX-609 fix some typos, pass an option  useState to the collaboration…

Atticus29:
… dialog and handle its setting to false in a useEffect call


src/components/cards/collaborations/CollaborationsDialog.jsx
```diff
@@ -109,14 +110,16 @@ export default function UserEditDialog({
           <Button
             display="primary"
             onClick={async () => {
+              if (setCollabDialogButtonClickLoading)
```
brmscheiner, line 113: if it's true then set it to true?

brmscheiner (approved):

PR #220, opened 2021-12-31
DEX-622 hide gps if gps data array is all null

Atticus29:


src/utils/formatters.js
```diff
@@ -1,6 +1,9 @@
 import { get, round, find, some, map } from 'lodash-es';
 import { format } from 'date-fns';
 
+export const isNull = inputArr =>
+  inputArr.join().replace(/,/g, '').length === 0;
```
brmscheiner, line 5: why this instead of a simple check for empty strings or null values?

Atticus29: No real reason. I just felt like being an idiot. Fixed?

brmscheiner (approved):

PR #221, opened 2022-01-04
Tweaks

brmscheiner:
- fix sightings page crash (GPS hotfix)
- remove post sighting functionality (always post AGS now)
- fix annotation menus 
- support domId prop on text elements with an example


Atticus29 (approved):
For the first time ever, I have no questions nor comments about this.

PR #227, opened 2022-01-11
Time specificity UI

brmscheiner:


src/components/fields/view/SpecifiedTimeViewer.jsx
```diff
@@ -0,0 +1,40 @@
+import React from 'react';
+import { get } from 'lodash-es';
+
+import { formatDateCustom } from '../../../utils/formatters';
+import Text from '../../Text';
+
+const formatMap = {
```
Atticus29, line 7: It seems like you might could DRY this up by filtering on precisionMap from SpecifiedTimeEditor? As in, perhaps they could both rely on this similar object in a utils files instead?

src/constants/fieldTypesNew.js
```diff
@@ -9,6 +9,7 @@ const fieldTypes = {
   latlong: 'latlong', // [float, float] defines [lat, lng]
   date: 'date',
   daterange: 'daterange', // [date, date]
+  specifiedTime: 'specifiedTime', // { time: string, timePrecision: string }
```
Atticus29, line 12: This file name tho.

Atticus29 (approved):
Approved with a few comments.

src/constants/fieldTypesNew.js
brmscheiner: yeahh i still need to delete the old one!!!

brmscheiner: it's on my todo list...

src/components/fields/view/SpecifiedTimeViewer.jsx
brmscheiner: good call - done!

PR #228, opened 2022-01-11
DEX-643 create pre-committed sightings table on user landing page and…

Atticus29:
… fix bugs with the sighting table

### DIFF
--- a/src/components/UserProfile.jsx
+++ b/src/components/UserProfile.jsx
@@ -133,35 +133,21 @@ export default function UserProfile({
         </CardContainer>
         <CardContainer>
           <SightingsCard
-            title={
-              someoneElse ? (
-                <FormattedMessage
-                  id="USERS_UNPROCESSED_AGS"
-                  values={{ name }}
-                />
-              ) : (
-                <FormattedMessage id="PENDING_SIGHTINGS" />
-              )
+            titleId={
+              someoneElse
+                ? 'USERS_UNPROCESSED_AGS'
+                : 'PENDING_SIGHTINGS'
             }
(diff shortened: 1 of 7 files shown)


src/components/UserProfile.jsx
```diff
@@ -128,6 +132,22 @@ export default function UserProfile({
...
+            title={
+              someoneElse ? (
+                <FormattedMessage
+                  id="USERS_UNPROCESSED_AGS"
+                  values={{ name }}
+                />
+              ) : (
+                <FormattedMessage id="PENDING_SIGHTINGS" />
+              )
+            }
```
brmscheiner, line 145: there is also a titleId prop. it's probably simpler to use that, though we might need to add support for values

src/components/UserProfile.jsx
```diff
@@ -128,6 +132,22 @@ export default function UserProfile({
...
+            columns={['individual', 'date', 'location', 'actions']}
+            sightings={agsData ? agsData : []}
```
brmscheiner, line 147: `sightings={agsData || []}`

src/components/UserProfile.jsx
```diff
@@ -141,7 +161,7 @@ export default function UserProfile({
             }
             columns={['individual', 'date', 'location', 'actions']}
             hideSubmitted
-            sightings={get(sightingsData, 'sightings', [])}
+            sightings={sightingsData ? sightingsData : []}
```
brmscheiner, line 164: `sightings={sightingsData || []}`

brmscheiner (approved):

PR #229, opened 2022-01-12
Support failed state in status card

brmscheiner:
image


src/pages/sighting/statusCard/StatusCard.jsx
```diff
@@ -44,35 +44,33 @@ export default function StatusCard({ sightingData }) {
...
+    curationComplete && sightingStage !== 'reviewed';
 
   let detectionStage = stages.waiting;
   let curationStage = stages.waiting;
   let matchingStage = stages.waiting;
 
-  if (detectionStartTime) detectionStage = stages.current;
+  if (detectionStartTime) {
+    detectionStage =
+      sightingStage === 'failed' ? stages.failed : stages.current;
```
Atticus29, line 63: More a cosmetic comment than anything:
sightingStage here seems like it will be easily confused with stages.
Maybe calling it sightingStatus instead would be more informative?
I dunno. Feel free to disregard. I had to read this a few times, though.

Atticus29: Also, `sightingStage === 'failed' ? stages.failed : stages.current;` gets repeated several times below. 
I sense an opportunity to DRY.

Atticus29 (approved):
LGTM. Just cosmetic suggestions.

src/pages/sighting/statusCard/StatusCard.jsx
brmscheiner: changed the name! not gonna bother with DRY right now since it's just one line repeated a couple times, but if it gets worse i'll address that too

PR #230, opened 2022-01-12
Dex 652 time and time specificity bulk import

Atticus29:


src/components/fields/edit/SpecifiedTimeEditor.jsx
```diff
@@ -18,14 +18,7 @@ import Text from '../../Text';
...
 export default function SpecifiedTimeEditor(props) {
-  const {
-    schema,
-    value,
-    onChange,
-    width = 280,
-    minimalLabels = false,
-    ...rest
-  } = props;
+  const { schema, value, onChange, width = 280, ...rest } = props;
```
brmscheiner, line 21: Okay, I think I remember now. If you remove `minimalLabels` it becomes part of `rest` and gets stuck on the DOM element. So deconstructing the variable prevents a console warning

src/components/fields/view/LocationIdViewer.jsx
```diff
@@ -4,7 +4,13 @@ import { flatten, get } from 'lodash-es';
 import Text from '../../Text';
 
 function findLocationObject(searchId, locationList) {
-  if (locationList === [] || !locationList || !searchId) return null;
+  if (
+    locationList === [] ||
+    !locationList ||
+    !searchId ||
+    locationList.length === 0
```
brmscheiner, line 11: Is there a difference between `locationList === []` and `locationList.length === 0`?

src/models/encounter/useEncounterFieldSchemas.js
```diff
@@ -91,6 +91,12 @@ export default function useSightingFieldSchemas() {
           category: defaultEncounterCategories.animal.name,
           hideOnBasicReport: true,
         }),
+        createFieldSchema(fieldTypes.date, {
```
brmscheiner, line 94: I think this is a mistake.. `specifiedTime` is already there in the same file

PR #233, opened 2022-01-13
DEX-633 and DEX-634 include customFields in the AGS and sightings pat…

Atticus29:
…ch calls


Atticus29:
> do they work now?!! hahaha

Somehow, that seemed sufficient lol!

PR #234, opened 2022-01-13
Create and delete AGS Encounters

brmscheiner:


src/components/cards/MetadataCardNew.jsx
```diff
@@ -11,6 +11,14 @@ import Button from '../Button';
 import Text from '../Text';
 import Card from './Card';
 
+function fieldValueGood(field) {
+  const value = field?.value;
+  if (value === null || value === undefined) return false;
```
Atticus29, line 16: What if field.value === ''?

Atticus29 (approved):
LGTM. Just one comment.

src/components/cards/MetadataCardNew.jsx
brmscheiner: good point

PR #235, opened 2022-01-14
DEX-664 fix the pins moving with map zooms, fix dead links in map mod…

Atticus29:
…e for pending sightings

### DIFF
--- a/src/components/maps/Marker.jsx
+++ b/src/components/maps/Marker.jsx
@@ -26,7 +26,7 @@ export default function Marker({
           style={{ backgroundColor: 'white', width: 'max-content' }}
         >
           <ButtonLink
-            href={`/${get(entry, 'linkPath')}/${get(entry, 'guid')}`}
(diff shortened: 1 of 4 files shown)


src/components/maps/Marker.jsx
```diff
@@ -26,7 +26,7 @@ export default function Marker({
           style={{ backgroundColor: 'white', width: 'max-content' }}
         >
           <ButtonLink
-            href={`/sightings/${get(entry, 'guid')}`}
+            href={`/${get(entry, 'linkPath')}/${get(entry, 'guid')}`}
```
brmscheiner, line 29: let's use `entry?.linkPath` style when we can remember to do it!

brmscheiner (approved):

src/components/maps/Marker.jsx
Atticus29: Done

PR #248, opened 2022-02-02
Qa fixes

brmscheiner:
- Use proper indicator of request success for POST site settings 
- Set height in photo detail UIs

### DIFF
--- a/src/pages/sighting/AnnotationDetail.jsx
+++ b/src/pages/sighting/AnnotationDetail.jsx
@@ -18,6 +18,8 @@ import useAddKeyword from '../../models/keyword/useAddKeyword';
 import Keywords from './Keywords';
 import MorePhotoMenu from './MorePhotoMenu';
 
+const imageMaxHeight = '80vh';
+
 export default function AnnotationDetail({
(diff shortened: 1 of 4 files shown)


src/pages/sighting/AnnotationDetail.jsx
```diff
@@ -18,6 +18,8 @@ import useAddKeyword from '../../models/keyword/useAddKeyword';
 import Keywords from './Keywords';
 import MorePhotoMenu from './MorePhotoMenu';
 
+const imageMaxHeight = '80vh';
```
Atticus29, line 21: This is the first I've seen of viewPort height. Why?

Atticus29 (approved):
LGTM

src/pages/sighting/AnnotationDetail.jsx
brmscheiner: well, normally we try to stick to a 4px grid system. but in this case my concern is that the photo will grow to be larger than the viewport. so maybe best to express measurements in terms of the viewport for this?

PR #249, opened 2022-02-02
Include Frontend branch and hash in site status page

brmscheiner:
After:
image


brmscheiner:
@Atticus29 don't worry about reviewing this one, at this point I think I have it working on `npm start` and `npm build` but some effort is needed on the houston side to get it building in docker (`docker-entrypoint.sh` doesn't use either of these commands). When @mmulich has things working on that side this PR can be merged

PR #252, opened 2022-02-03
DEX-712 change all references to passed_message into message

Atticus29:


hwindsor (approved):
It may have not been clear but In the Jira I said to remove all uses of 'passed_message' in the collaborations front_end code ONLY. Not POST of Asset groups, Annotations or Individuals. 
But looking at the Houston code in more detail, Asset groups and Annotations POST does not use passed_message but Individual POST does.
So if this change goes in, then Individual POST request will not have the correct error message on failure. 
But I have created a Jira to sort all of this out in Houston (714) so I think your change should go in as it is.

brmscheiner (approved):
@Atticus29 lgtm but I'll let you decide whether to merge before JIRA 714 goes through

hwindsor:
> @Atticus29 lgtm but I'll let you decide whether to merge before JIRA 714 goes through

This fixes a huge amount of issues in the error hhandling from houston so I'd say that it should go in first. The only thing that will be lost with this is the error response for Individual POST which will be fixed by 714 which is also in PR. 
I'd say this one should go in before 714.

PR #254, opened 2022-02-04
Individuals sighting table using cell renderers

brmscheiner:


src/components/cards/SightingsCard.jsx
```diff
@@ -73,51 +57,15 @@ export default function SightingsCard({
...
-            >
-              {get(
-                sightingObj,
-                'formattedLocation',
-                intl.formatMessage({ id: 'LOCATION_NAME_MISSING' }),
-              )}
-            </Button>
-          );
-        },
+        cellRenderer: cellRendererTypes.location,
```
brmscheiner, line 68: This PR adds the ability to specify a "cell renderer" when setting up a table. It also provides the first handful of renderers, which are meant to abstract logic that was being repeated in multiple tables. This diff here is the goal: replace complicated customBodyRender functions that are repeated in multiple places with a simple config string

src/components/dataDisplays/CollapsibleRow.jsx
```diff
@@ -42,11 +40,22 @@ export default function CollabsibleRow({
...
+            c,
+            'options.cellRendererProps',
+            {},
+          );
+          const customCellRenderer = get(
             c,
             'options.customBodyRender',
-            defaultCellRenderer,
           );
+          const RequestedCellRenderer =
```
Atticus29, line 57: why the capital letter?

src/components/dataDisplays/CollapsibleRow.jsx
```diff
@@ -57,7 +66,15 @@ export default function CollabsibleRow({
                 ...cellStyles,
               }}
             >
-              {cellRenderer(cellValue, datum)}
+              {customCellRenderer ? (
+                customCellRenderer(cellValue, datum)
+              ) : (
+                <RequestedCellRenderer
```
Atticus29, line 72: There's something that I might not understand here. How is this a component??

src/components/dataDisplays/CollapsibleRow.jsx
```diff
@@ -8,16 +8,14 @@ import DownIcon from '@material-ui/icons/KeyboardArrowDown';
 import UpIcon from '@material-ui/icons/KeyboardArrowUp';
 import Collapse from '@material-ui/core/Collapse';
 
+import { cellRenderers } from './cellRenderers';
```
Atticus29, line 11: Did not know that you could import multiple files this way. Very cool.

src/components/dataDisplays/CollapsibleRow.jsx
Atticus29: Nvm. I get it now. The renderers imported from the renderers directory are the components.

src/components/dataDisplays/CollapsibleRow.jsx
Atticus29: Ignore. I get it now.

src/components/dataDisplays/cellRenderers/index.js
```diff
@@ -0,0 +1,18 @@
+import DefaultRenderer from './DefaultRenderer';
```
Atticus29, line 1: I don't understand why this file needs to be an index.js. Is that important?

Atticus29: Maybe you _can't_ import multiple files as I stated above. It's just that importing a directory looks for index.js if no file is named explicitly?

Atticus29 (approved):
LGTM. Mostly just educational questions I've asked here.

src/components/dataDisplays/cellRenderers/index.js
brmscheiner: Exactly!

PR #255, opened 2022-02-04
Dex 663 notification uis for different notification types

Atticus29:


Atticus29:
Notes:
this uses a new notification schema. From said schema, a dev. can indicate title (titleId), whether or not to display a "View" button (displayViewMoreInfoButton) from the notification pane next to the abbreviated message (notificationMessage)  [note that displayViewMoreInfoButton plugs directly in to the display attribute; hence, the 'none'], provide a more detailed message in the notification dialog (moreDetailedDescription), and provide a list of available buttons from within the dialog box (availableButtons), e.g. grant, decline, revoke, and request.

Known issues/shortcomings: 

1. If any of the grant, decline, etc. buttons fail, there is no way to dismiss the error message and that error message will persist upon additional opened dialogs from notifications.jsx. This will likely involve a minor refactor of the custom hooks for grant, decline, etc. to allow us to setError to null upon dismissal locally.

2. Revoke and Request buttons have not been fleshed out yet. Indeed, I want confirmation that those functionalities would be desired from the notifications; I just assumed so. 

Since especially this second issue will likely be a little involved, I want to create separate tickets for these issues.

Atticus29:
Follow up: created a separate ticket for the issues above: https://wildme.atlassian.net/browse/DEX-727

locale/en.json
```diff
@@ -951,7 +953,24 @@
...
-  "TIME_SINCE": "{timeSince} ago"
+  "A_COLLABORATION_WAS_CREATED_ON_YOUR_BEHALF_MORE_DETAILED": "A collaboration was created on your behalf between {user1Name} and {user2Name}",
+  "TIME_SINCE": "{timeSince} ago",
+  "COLLABORATION_APPROVED": "Your collaboration with {userName} has been approved",
+  "COLLABORATION_APPROVED_TITLE": "Collaboration approved",
+  "COLLABORATION_REVOKE_BRIEF": "{userName} revoked your collaboration request",
+  "EDIT_COLLABORATION_APPROVED": "{userName} approved your collaboration edit request",
+  "EDIT_COLLABORATION_REVOKED": "{userName} revoked edit privileges for your collaboration with them",
+  "COLLABORATION_REVOKED_BY_MANAGER": "A collaboration was revoked by a user manager",
+  "INDIVIDUAL_MERGE_REQUEST_MESSAGE": "{userName} requested that an individual in your data set be merged with another individual in the database",
```
brmscheiner, line 966: For this one is there any way to access the names of the individuals to be merged? I was thinking something like `Jake Holland wants to merge "Stegosaurus" with "Big Beefy".`

src/components/AuthenticatedAppHeader/NotificationsPane.jsx
```diff
@@ -116,7 +117,11 @@ export default function NotificationsPane({
                       await markRead(get(notification, 'guid'));
                       refreshNotifications();
                     }}
-                    style={{ maxHeight: 40 }}
+                    style={{
+                      maxHeight: 40,
+                      display:
+                        currentNotificationSchema?.displayViewMoreInfoButton,
```
brmscheiner, line 123: While I think the setup for the display text is good, I'm not sure this is the right way to do the buttons. If a view button isn't part of the notification at all, it shouldn't be rendered to the DOM. (Here it is rendered with `display: none`.)

Instead it could be like this:`{displayViewMoreInfoButton && <Button id="VIEW" ... />}`

src/constants/notificationSchema.js
```diff
@@ -0,0 +1,76 @@
...
+    moreDetailedDescription: 'COLLABORATION_VIEW_REQUEST_DESCRIPTION',
+    availableButtons: ['grant', 'decline'],
+    path: '/view_permission',
+  },
+  collaboration_approved: {
+    titleId: 'COLLABORATION_APPROVED_TITLE',
+    displayViewMoreInfoButton: 'none',
+    notificationMessage: 'COLLABORATION_APPROVED',
+    moreDetailedDescription: 'COLLABORATION_APPROVED',
+    availableButtons: ['revoke'],
```
brmscheiner, line 23: it's weird to define the list of possible buttons here, but have view be treated separately above. my recommendation would be to handle all the buttons together and come up with some sort of schema for defining them. for example 

```
availableButtons: [{
  id: 'approve-collaboration',
  labelId: 'APPROVE',
  display: 'primary',
  href: null,  // buttons would have links or onClick actions, but not both 
  onClick: async (notification) => ....
}]

```

brmscheiner (changes requested):
I think this is mostly good, but more work should be done on the way buttons are defined in the schema

src/components/AuthenticatedAppHeader/NotificationsPane.jsx
Atticus29: Moot because of below.

src/constants/notificationSchema.js
Atticus29: Ok I understand this one and the value of having all of the button prop logic live in notificationSchema.js.

However. Let's take the 'view' as an example. In NotificationsPane.jsx, I currently have:

```
{availableButtons.includes('view') && (
                    <Button
                      onClick={async () => {
                        setActiveCollaborationNotification(
                          notification,
                        );
                        await markRead(get(notification, 'guid'));
                        refreshNotifications();
                      }}
                      style={{ maxHeight: 40 }}
                      id="VIEW"
                    />
                  )}
```

Handling setActiveCollaborationNotification and refreshNotifications in notificationsSchema seems like it will require some gymnastics.

Similar issue for accessing onClose (a function passed into NotificationDetailsDialog as a prop) for 'decline' and 'grant'.

What do you think? More generally this is something I struggle with a lot: there's where I think the logic of these buttons, etc.  _ought_  to be written (in this case, arguably notificationSchema?) and then there's where _react seems to like them written_ (in this case, in the individual component files... e.g., NotificationDetailsDialog.jsx and NotificationsPane.jsx).

brmscheiner: Yeah i see where you're coming from. But the existing code will get messy once we have non-collaboration notifications and need to do something different instead of `setActiveCollaborationNotification`. Maybe the best solution is to do something similar to the cell renderer PR that just went in (or the way we handle fields) and define a separate notification component for each notification type.

Atticus29: Yeah, that's kinda along the lines of what I was thinking, too. In this case, I will pivot to DEX-719 to get a better handle on that concept.

src/components/dialogs/notificationDialogUtils/NotificationCollaborationGrantAccessButton.jsx
```diff
@@ -0,0 +1,66 @@
...
+import { useIntl } from 'react-intl';
+import { useQueryClient } from 'react-query';
+import { get } from 'lodash-es';
+import DialogContent from '@material-ui/core/DialogContent';
+
+import Button from '../../Button';
+import CustomAlert from '../../Alert';
+import usePatchCollaboration from '../../../models/collaboration/usePatchCollaboration';
+import queryKeys from '../../../constants/queryKeys';
+// import Text from '../../Text';
```
brmscheiner, line 11: could remove this

src/components/dialogs/notificationDialogUtils/NotificationCollaborationGrantAccessButton.jsx
```diff
@@ -0,0 +1,66 @@
...
+    isError,
+  } = usePatchCollaboration();
+  const collaborationId = get(notification, [
+    'message_values',
+    'collaboration_guid',
+  ]);
+  return (
+    // <Text>Hi there</Text>
+    <DialogContent>
+      {/* I know that this is a nested DialogContent. If you have better advice for how to handle the fact that isError now lives locally withing these button files instead of the in the NotificationDetailsDialog, I'm all ears */}
```
brmscheiner, line 33: why not just a div?

src/components/dialogs/notificationDialogUtils/NotificationCollaborationGrantAccessButton.jsx
```diff
@@ -0,0 +1,66 @@
...
+    loading,
+    error,
+    isError,
+  } = usePatchCollaboration();
+  const collaborationId = get(notification, [
+    'message_values',
+    'collaboration_guid',
+  ]);
+  return (
+    // <Text>Hi there</Text>
```
brmscheiner, line 31: could remove this

src/components/dialogs/notificationDialogUtils/NotificationCollaborationGrantAccessButton.jsx
```diff
@@ -0,0 +1,66 @@
...
+          );
+          if (response?.status === 200) {
+            onCloseDialog();
+            queryClient.invalidateQueries(queryKeys.me);
+          }
+        }}
+        id="GRANT_ACCESS"
+      />
+      {isError && (
+        <CustomAlert
```
brmscheiner, line 56: i feel like this error will show up down by the buttons? that seems like a weird place for it, though i could be wrong

src/components/dialogs/notificationDialogUtils/index.js
```diff
@@ -0,0 +1,17 @@
...
+  grant: 'grant',
+  decline: 'decline',
+  revoke: 'revoke',
+  request: 'request',
+};
+
+export const notificationButtons = {
+  [notificationButtonTypes.view]: NotificationViewButton,
+  [notificationButtonTypes.grant]: NotificationCollaborationGrantAccessButton,
+  [notificationButtonTypes.decline]: NotificationDeclineCollaborationRequestButton,
```
brmscheiner, line 16: TBH i'm still not totally sold on this approach of treating the buttons separately from the notification content. i think the question of "where do we put the error that happens in the button component" is a symptom of having separated two things that can't be easily treated separately.

Atticus29: As in, you're envisioning the components in e.g., NotificationViewButton, etc. to be in the notificationSchema file?

src/components/dialogs/notificationDialogUtils/NotificationCollaborationGrantAccessButton.jsx
Atticus29: I tried it the other way and ended up moving the error here. I wish that I remembered why. I'll try it in the parent dialog file and see whether it works now.

src/components/dialogs/notificationDialogUtils/index.js
brmscheiner: In my vision each notification has its own file, content and buttons together. But if you are sure this is working I'm ok with that. It just confuses me a little

Atticus29: It is working. And I agree; feels a little overkill-y with the abstraction.

src/components/AuthenticatedAppHeader/NotificationsPane.jsx
```diff
@@ -82,43 +101,36 @@ export default function NotificationsPane({
...
-                      );
-                      await markRead(get(notification, 'guid'));
-                      refreshNotifications();
-                    }}
-                    style={{ maxHeight: 40 }}
-                    id="VIEW"
-                  />
+                  {get(
+                    currentNotificationSchema,
+                    'viewInNotificationPane',
```
Atticus29, line 112: I changed the viewButton back to a boolean in the schema. Since all of the other button stuff is no longer in the schema, and the view button is the only button that appears in the NotificationPane, this feels like a more natural way to handle it in the new setup.

brmscheiner: Sounds great!

brmscheiner: could use `currentNotificationSchema?.viewInNotificationPane` here instead if you want

brmscheiner: It's a little weird to chain `{condition && condition && <component />}` across multiple lines. you could do something like this in the function body instead 

`const showViewButton = currentNotificationSchema?.viewInNotificationPane && refreshNotifications !== undefined`

then in render `{showViewButton && ...`

src/components/dialogs/notificationDialogUtils/NotificationCollaborationEditRevokeDialog.jsx
```diff
@@ -0,0 +1,32 @@
...
+      onClose={onClose}
+      notification={notification}
+      titleId={get(currentNotificationSchema, 'titleId')}
+      moreDetailedDescription={get(
+        currentNotificationSchema,
+        'moreDetailedDescription',
+      )}
+      buttons={avaialableButtons}
+      isError={false}
+      error=""
```
brmscheiner, line 29: error = ""?

src/components/dialogs/notificationDialogUtils/NotificationCollaborationEditRevokeDialog.jsx
```diff
@@ -0,0 +1,32 @@
...
+  notification,
+  onClose,
+  open,
+}) {
+  const notificationType = notification?.message_type;
+  const currentNotificationSchema = get(
+    notificationSchema,
+    notificationType,
+  );
+  const avaialableButtons = [];
```
brmscheiner, line 16: typo

src/components/dialogs/notificationDialogUtils/NotificationCollaborationEditRevokeDialog.jsx
```diff
@@ -0,0 +1,32 @@
...
+      open={open}
+      onClose={onClose}
+      notification={notification}
+      titleId={get(currentNotificationSchema, 'titleId')}
+      moreDetailedDescription={get(
+        currentNotificationSchema,
+        'moreDetailedDescription',
+      )}
+      buttons={avaialableButtons}
+      isError={false}
```
brmscheiner, line 28: should this default to false?

src/components/dialogs/notificationDialogUtils/NotificationCollaborationManagerCreateDialog.jsx
```diff
@@ -0,0 +1,32 @@
...
+  notification,
+  onClose,
+  open,
+}) {
+  const notificationType = notification?.message_type;
+  const currentNotificationSchema = get(
+    notificationSchema,
+    notificationType,
+  );
+  const avaialableButtons = [];
```
brmscheiner, line 16: typo

src/components/dialogs/notificationDialogUtils/NotificationCollaborationManagerRevokeDialog.jsx
```diff
@@ -0,0 +1,32 @@
...
+  notification,
+  onClose,
+  open,
+}) {
+  const notificationType = notification?.message_type;
+  const currentNotificationSchema = get(
+    notificationSchema,
+    notificationType,
+  );
+  const avaialableButtons = [];
```
brmscheiner, line 16: typo

brmscheiner: actually i'm noticing there are several of these, and also several `error=""` below. just do a scan for them please and i'll stop sending you emails for each of them :)

src/components/dialogs/notificationDialogUtils/NotificationCollaborationEditRevokeDialog.jsx
Atticus29: So, for several of these dialogs, no buttons and therefore no api calls are required. I suppose I could pass null? Or have them default to null in notificationDialog? Same for the similar comments below.

src/components/dialogs/notificationDialogUtils/NotificationCollaborationEditRevokeDialog.jsx
Atticus29: Yes (see above).

src/components/dialogs/notificationDialogUtils/NotificationCollaborationManagerRevokeDialog.jsx
Atticus29: See above.

src/constants/notificationSchema.js
```diff
@@ -0,0 +1,71 @@
+export const notificationSchema = {
+  collaboration_manager_create: {
```
brmscheiner, line 2: maybe these property names should come from `notificationTypeNames` above? seems like they would need to be kept in sync manually this way

src/components/dialogs/notificationDialogUtils/NotificationCollaborationApprovedDialog.jsx
```diff
@@ -0,0 +1,66 @@
+import React from 'react';
```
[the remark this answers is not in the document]
brmscheiner: I see what you mean but i think this is good for now! In their current state there is a lot of repeated code, but we might find that the UIs for these notifications continue to diverge as the product develops. And if not, then let's DRY it up l8r!

brmscheiner (approved):
approved with comments!

src/components/dialogs/notificationDialogUtils/NotificationCollaborationEditRevokeDialog.jsx
brmscheiner: Exactly, just use default values

src/components/AuthenticatedAppHeader/NotificationsPane.jsx
Atticus29: Done

src/components/dialogs/notificationDialogUtils/NotificationCollaborationEditRevokeDialog.jsx
Atticus29: Done

src/components/dialogs/notificationDialogUtils/NotificationCollaborationEditRevokeDialog.jsx
Atticus29: Done

src/components/dialogs/notificationDialogUtils/NotificationCollaborationEditRevokeDialog.jsx
Atticus29: Done

src/components/dialogs/notificationDialogUtils/NotificationCollaborationManagerCreateDialog.jsx
Atticus29: Done

src/components/dialogs/notificationDialogUtils/NotificationCollaborationManagerRevokeDialog.jsx
Atticus29: Done

PR #258, opened 2022-02-10
Cleanup

brmscheiner:


src/components/cards/MetadataCard.jsx
```diff
@@ -1,26 +1,56 @@
...
+  let metadataToDisplay = metadata;
+  if (!showDefaultValues) {
+    metadataToDisplay = metadata.filter(
+      field =>
+        JSON.stringify(field.value) !==
+        JSON.stringify(field.defaultValue),
+    );
+  }
+
+  metadataToDisplay = metadataToDisplay.filter(
```
Atticus29, line 41: I think we want to avoid the metadataToDisplay = metadataToDisplay pattern, right? Maybe combine the filters?

src/components/cards/MetadataCard.jsx
```diff
@@ -34,38 +64,57 @@ export default function MetadataCard({
...
-              <ListItemText
-                primary={
-                  <Text
-                    component="span"
-                    variant="caption"
-                    id={datum.titleId}
+        {metadata
+          ? metadataToDisplay.map(field => {
+              const viewComponentProps =
+                field.viewComponentProps || {};
```
Atticus29, line 70: Should all of these field. accessors be field?.  ?

Atticus29: I guess not since we know field exists. I'm just (irrationally?) worried that one of these will not work.

src/components/cards/MetadataCard.jsx
```diff
@@ -34,38 +64,57 @@ export default function MetadataCard({
...
-                  )
-                }
-              />
-            </ListItem>
-          );
-        })}
+                </ListItem>
+              );
+            })
+          : range(6).map(i => (
```
Atticus29, line 101: Why 6?

Atticus29: Oh. Only if it's completely empty, right? Then populate with six skeleton items?

src/pages/sighting/EditSightingMetadata.jsx
```diff
@@ -141,12 +141,12 @@ export default function EditSightingMetadata({
           loading={loading}
           display="primary"
           onClick={async () => {
+            const properties = { ...defaultFieldValues };
```
Atticus29, line 144: How is this different from const properties = defaultFieldValues?

(Unless perhaps defaultFieldValues is an array, in which case I take back the question.)

Atticus29 (approved):
A few minor suggestions.

src/components/cards/MetadataCard.jsx
brmscheiner: yes

src/pages/sighting/EditSightingMetadata.jsx
brmscheiner: this creates a shallow copy of the original object, which is helpful for preventing certain bugs as properties gets edited

src/components/cards/MetadataCard.jsx
brmscheiner: good call

src/components/cards/MetadataCard.jsx
brmscheiner: yeah fair enough. couldn't hurt

PR #259, opened 2022-02-11
Three small tickets

brmscheiner:
- Fix sighting header
- Disable react-query devtools 
- Fix enter functionality on annotation detail page 
- Use hook to abstract enter key functionality everywhere on site


Atticus29 (approved):
LGTM

PR #261, opened 2022-02-16
Fix individual sightings table

brmscheiner:
Problem was sighting properties were not getting copied to new encounters so table had the wrong data.


Atticus29 (approved):
LGTM

PR #263, opened 2022-02-17
Dex 708 sigthing pending sighting loading

Atticus29:


Atticus29:
Ok... I had to create cards in user.jsx that just had linear progress bars because linear progress bars in DataDisplay that would _normally_ display loading status were themselves throttled by the lengthy api calls in the case of Rosemary. This was my solution.

src/components/UserProfile.jsx
```diff
@@ -132,23 +155,41 @@ export default function UserProfile({
...
-            columns={['individual', 'date', 'location', 'actions']}
-            hideSubmitted
-            sightings={sightingsData || []}
-          />
+          {localAgsLoading && (
+            <Card titleId={'PENDING_SIGHTINGS'}>
+              <LinearProgress style={{ margin: '16px 32px' }} />
+            </Card>
+          )}
+          {!localAgsLoading && (
```
brmscheiner, line 163: you could use a ternary here 
```
{localSightings ? (...) : (...)}
```

src/components/UserProfile.jsx
```diff
@@ -132,23 +155,41 @@ export default function UserProfile({
...
+            </Card>
+          )}
+          {!localSightingsLoading && (
+            <SightingsCard
+              id="sightings-card"
+              titleId={someoneElse ? 'USERS_SIGHTINGS' : 'SIGHTINGS'}
+              columns={['individual', 'date', 'location', 'actions']}
+              hideSubmitted
+              sightings={sightingsData || []}
+              loading={localSightingsLoading}
```
brmscheiner, line 190: this would be false always right?

src/components/UserProfile.jsx
```diff
@@ -132,23 +155,41 @@ export default function UserProfile({
...
+              titleId={
+                someoneElse
+                  ? 'USERS_UNPROCESSED_AGS'
+                  : 'PENDING_SIGHTINGS'
+              }
+              columns={['individual', 'date', 'location', 'actions']}
+              sightings={agsData || []}
+              linkPath="pending-sightings"
+              noSightingsMsg="NO_PENDING_SIGHTINGS"
+              loading={localAgsLoading}
```
brmscheiner, line 175: this would be false always right?

src/components/UserProfile.jsx
```diff
@@ -28,12 +30,33 @@ export default function UserProfile({
...
+  );
+  useEffect(
+    () => {
+      setLocalAgsLoading(agsLoading);
+    },
+    [agsLoading],
+  );
+  useEffect(
+    () => {
+      setLocalSightingsLoading(sightingsLoading);
```
brmscheiner, line 57: what I am trying to figure out is: in what situation does `localSightingsLoading !== sightingsLoading`?

src/components/UserProfile.jsx
Atticus29: I believe you're right. I thought of just dropping the loading prop entirely from the SightingsCard components, but I figured that this would future-proof things a little better if we eventually drop the other Card element after we speed up the AGS get api call (ticket https://wildme.atlassian.net/browse/DEX-786). Happy to drop  loading={localSightingsLoading} if you agree that's the right call.

src/components/UserProfile.jsx
Atticus29: See above.

src/components/UserProfile.jsx
Atticus29: I don't think that situation exists either. Just some overkill.

src/components/UserProfile.jsx
Atticus29: TY

brmscheiner (approved):

PR #264, opened 2022-02-18
React query abstraction

brmscheiner:
I skipped over three get hooks while doing this migration:
- `useServerStatus`, which is part of Emily's original open source contribution. For one of her first tasks I plan to have her migrate to the new standard 
- `useQueryIndividuals` and `useFilterIndividuals`. technically these are POSTs, but they act exactly like GETs and that confuses me. after confirmation with the backend team that it should be this way, i'll probably extend `useGet` so you can override it and send a different kind of request with the same logic 

`useGet` supports two methods for backwards compatibility but we should migrate usages of these ASAP 
- `setError`: i'm pretty sure we should be using `clearError` instead everywhere. i can't remember a good reason why the error should ever be changed to a different error 
- `refresh`: eventually we should support react-query invalidation instead, but that will have to wait until we perform a similar exercise for mutations as well


src/constants/queryKeys.js
```diff
@@ -22,3 +27,19 @@ export function getAGSQueryKey(id) {
 export function getIndividualQueryKey(id) {
   return ['individual', id];
 }
+
+export function getUserQueryKey(id) {
```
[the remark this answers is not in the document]
Atticus29: Nvm. Makes sense. I forgot about compound query keys.

src/components/UserProfile.jsx
```diff
@@ -155,57 +136,44 @@ export default function UserProfile({
...
+                ? intl.formatMessage(
+                    { id: 'USERS_SIGHTINGS' },
+                    { name },
+                  )
+                : intl.formatMessage({ id: 'SIGHTINGS' })
+            }
+            columns={['individual', 'date', 'location', 'actions']}
+            hideSubmitted
+            sightings={sightingsData || []}
+            loading={sightingsLoading}
```
Atticus29, line 172: I worry that this introduces the regression of:
> first DataDisplay was only getting rendered when there was data, so after loading was finished, so no bar was being loaded because of that

src/models/assetGroup/useAssetGroup.js
```diff
@@ -1,52 +1,9 @@
...
-      if (assetGroupId) fetchAssetGroupData();
-    },
-    [assetGroupId, refreshCount],
-  );
-
-  return { data, statusCode, loading, error, refresh };
+  return useGet({
+    queryKey: getAssetGroupQueryKey(assetGroupId),
+    url: `/asset_groups/${assetGroupId}`,
+  });
```
Atticus29, line 8: Chef's kiss. So much DRYer.

src/models/assetGroupSighting/useAssetGroupSighting.js
```diff
@@ -0,0 +1,13 @@
+import { getAGSQueryKey } from '../../constants/queryKeys';
```
Atticus29, line 1: IMHO maybe we should start being more consistent with our naming conventions of these custom hooks. Do we include, "Get" in the file name? As in, useGetAssetGroupSighting, or not?

src/models/users/useGetMe.js
```diff
@@ -1,39 +1,23 @@
...
 
 export default function useGetMe() {
-  const result = useQuery(
-    queryKeys.me,
-    async () => {
-      const response = await axios.request({
-        url: `${__houston_url__}/api/v1/users/me`,
-        method: 'get',
+  function onSuccess(response) {
+    if (!__DEV__) {
```
Atticus29, line 8: ?

Atticus29: I don't understand what __DEV__ is or where it's coming from. I also get the eery feeling that I have asked this exact question before.

Atticus29 (commented):
Mostly questions and minor stuff.

src/hooks/useGet.js
```diff
@@ -0,0 +1,77 @@
...
+        method: 'get',
+        data,
+      });
+      const status = response?.status;
+      setStatusCode(status);
+      if (status === 200) onSuccess(response);
+      return response;
+    },
+    {
+      staleTime: Infinity,
```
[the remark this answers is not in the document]
brmscheiner: Yeah, that's what I was thinking. Usually in our app data expires because of an action, not because a certain amount of time has passed. The only exception I could think of was `GET ags/sighting`, so I figure better to use this as the dfault and override in that case.

src/models/assetGroupSighting/useAssetGroupSighting.js
brmscheiner: this isn't a custom hook, it's a query key getter. all custom hooks should begin with "use" for sure! now whether "get" is the best prefix for query keys... i guess it could also be derive? but for me get is ok too

src/models/users/useGetMe.js
brmscheiner: this is a build time variable that is set up in webpack. if you're running the app in its dev environment `__DEV__ === true`, if you are running the production build `__DEV__ ===  false`

src/components/UserProfile.jsx
brmscheiner: I tested the fix! Though please feel free to double-check me

src/hooks/useGet.js
```diff
@@ -0,0 +1,77 @@
...
+    : url;
+  const result = useQuery(
+    queryKey,
+    async () => {
+      const response = await axios.request({
+        url: apiUrl,
+        method: 'get',
+        data,
+      });
+      const status = response?.status;
```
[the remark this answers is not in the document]
brmscheiner: `react-query` status is indeed a string enum with loading/idle/etc. but this is the _axios_ status!

src/hooks/useGet.js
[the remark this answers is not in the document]
Atticus29: How would one override this??

src/models/assetGroupSighting/useAssetGroupSighting.js
Atticus29: Sorry... iirc, this comment was for that line, but rather for the file itself, which is named useAssetGroupSighting.js...

src/models/users/useGetMe.js
Atticus29: Thanks!

src/components/UserProfile.jsx
Atticus29: Hm. Do you happen to remember whether you tested it before or after updating houston when Jon fixed the speed issue?
https://wildme.atlassian.net/browse/DEX-786, which was merged four days ago.

brmscheiner: I think I probably tested after that was fixed. But I did use the network devtools to simulate very slow connections

src/hooks/useGet.js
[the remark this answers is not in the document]
brmscheiner: something like `queryOptions: { staleTime: 5000 }` would do the trick. Because queryOptions are spread onto the end so they override the defaults

src/models/assetGroupSighting/useAssetGroupSighting.js
brmscheiner: oh yes, we could definitely rename this to `useGetAssetGroupSighting` or `useGetAGS`

src/hooks/useGet.js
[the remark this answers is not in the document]
Atticus29: Very cool. Good to know. Thank you.

PR #266, opened 2022-02-23
DEX-727 add revoke access button and realize the request button for n…

Atticus29:
…otifications is unnecessary for MVP

Note that this PR is into DEX-663 rather than develop.


brmscheiner (approved):
i think it's fine to merge this into the other PR

PR #272, opened 2022-03-04
DEX-743 disambiguate users with same full name in collaboration form …

Atticus29:
…autocompletes

### DIFF
--- a/src/pages/collaborations/collaborationManagementForm.jsx
+++ b/src/pages/collaborations/collaborationManagementForm.jsx
@@ -56,15 +56,21 @@ export default function CollaborationManagementForm({
           getOptionLabel={option => {
             const name = get(option, 'full_name', null);
             const email = get(option, 'email', null);
-            if (name) return name + ' (' + email + ')';
-            return (
-              intl.formatMessage({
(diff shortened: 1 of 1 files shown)


src/pages/collaborations/collaborationManagementForm.jsx
```diff
@@ -55,8 +55,8 @@ export default function CollaborationManagementForm({
           }
           getOptionLabel={option => {
             const name = get(option, 'full_name', null);
-            if (name) return name;
             const email = get(option, 'email', null);
+            if (name) return name + ' (' + email + ')';
```
brmscheiner, line 59: is email guaranteed to be there? if not you could wind up with things like `Peter ()`

brmscheiner (approved):
approved with question

src/pages/collaborations/collaborationManagementForm.jsx
Atticus29: Good point. If a user doesn't have an email address, though, even for migrated data, we have a bigger problem. But I'll fix.

Emily-Ke (approved):

PR #273, opened 2022-03-08
DEX-754 comment out delete photograph option for now

Atticus29:


brmscheiner (approved):

Emily-Ke (approved):
Approved, but why not just delete it? We have the file history if we need to restore it.

Atticus29:
@Emily-Ke you make a good point about the commenting. This is a feature that we will almost assuredly be bringing back post-MVP. I've seen Ben do similar elsewhere for similar cases and am just following the pattern. I see value in both - it's annoying to see commented-out code that could potentially become cruft. However, it's also annoying to comb through the history to retrieve old code. I'm kind of agnostic on this one.

PR #275, opened 2022-03-09
DEX-802 remove terms and conditions for authorized users

Emily-Ke:
- Removes the terms and conditions checkbox for authorized users:
  - Removes all terms and conditions logic from the `BulkReportForm` because that form is only visible to authorized users.
  - Updates logic to conditionally require the terms and conditions acknowledgment in the `ReportForm`.
- Replaces the temporary terms and conditions text with the actual copy.

**Potential issue**: On some screen sizes, it is not obvious that the Codex Terms of Use continue beyond the bottom of the visible dialog:
dex-802-dialog2

### DIFF
--- a/src/components/TextContainers.jsx
+++ b/src/components/TextContainers.jsx
@@ -1,0 +1,60 @@
+import React from 'react';
+import Typography from '@material-ui/core/Typography';
+
+export function Header1({ children, ...rest }) {
+  return (
+    <Typography component="h1" variant="h4" gutterBottom {...rest}>
+      {children}
+    </Typography>
+  );
+}
+
+export function Header2({ children, ...rest }) {
+  return (
+    <Typography component="h2" variant="h5" gutterBottom {...rest}>
+      {children}
+    </Typography>
+  );
+}
+
+export function Header3({ children, ...rest }) {
+  return (
+    <Typography component="h3" variant="h6" gutterBottom {...rest}>
+      {children}
+    </Typography>
+  );
+}
+
+export function Header4({ children, ...rest }) {
+  return (
+    <Typography
+      component="h4"
+      variant="subtitle2"
+      gutterBottom
+      style={{ fontStyle: 'italic', fontWeight: 'bold' }}
+      {...rest}
+    >
+      {children}
+    </Typography>
+  );
+}
+
+export function Paragraph({ children, ...rest }) {
+  return (
+    <Typography paragraph {...rest}>
+      {children}
+    </Typography>
+  );
+}
+
+export function ListItem({ children, ...rest }) {
+  return (
+    <Typography component="li" {...rest}>
+      {children}
+    </Typography>
+  );
+}
+
+export function Definition({ children }) {
+  return <dfn style={{ fontStyle: 'normal' }}>{children}</dfn>;
+}
--- a/src/components/report/TermsAndConditionsDialog.jsx
+++ b/src/components/report/TermsAndConditionsDialog.jsx
@@ -1,7 +1,6 @@
 import React from 'react';
 import DialogContent from '@material-ui/core/DialogContent';
 import Typography from '@material-ui/core/Typography';
-import { makeStyles } from '@material-ui/core/styles';
 import StandardDialog from '../StandardDialog';
 
 function H1({ children, ...rest }) {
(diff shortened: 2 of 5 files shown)


src/pages/reportSighting/ReportForm.jsx
```diff
@@ -95,7 +96,7 @@ export default function ReportForm({
     schema => schema.customField,
   );
 
-  const [acceptedTerms, setAcceptedTerms] = useState(false);
+  const [acceptedTerms, setAcceptedTerms] = useState(authenticated);
```
Atticus29, line 99: Clever!

src/pages/reportSighting/ReportForm.jsx
```diff
@@ -214,7 +215,7 @@ export default function ReportForm({
           />
         </>
       )}
-      {sightingType && (
+      {sightingType && !authenticated && (
```
Atticus29, line 218: @brmscheiner has made the recommendation to me before to handle this logic elsewhere such that we only have one boolean here before the component render.
So, maybe something like 
```
const sightingTypeButNotAuthenticated  = sightingType && !authenticated;
...
{sightingTypeButNotAuthenticated && ( component logic)}
```

src/components/report/TermsAndConditionsDialog.jsx
```diff
@@ -1,24 +1,786 @@
 import React from 'react';
 import DialogContent from '@material-ui/core/DialogContent';
-import Text from '../Text';
+import Typography from '@material-ui/core/Typography';
+import { makeStyles } from '@material-ui/core/styles';
 import StandardDialog from '../StandardDialog';
 
+function H1({ children, ...rest }) {
```
Atticus29, line 7: I wonder whether a lot of these could be put in some kind of component utils file somewhere so that they're move easily accessible for other components as well.

@brmscheiner , thoughts?

src/components/report/TermsAndConditionsDialog.jsx
```diff
@@ -1,24 +1,786 @@
...
+
+function ListItem({ children, ...rest }) {
+  return (
+    <Typography component="li" {...rest}>
+      {children}
+    </Typography>
+  );
+}
+
+const useStyles = makeStyles(() => ({
```
Atticus29, line 61: TIL makeStyles!.... which looks to be deprecated in v5 of material?

Also, I think @brmscheiner strongly prefers inline styling.

Atticus29 (commented):
Generally looks great!
Left some comments. Will await Ben's opinion.

src/components/report/TermsAndConditionsDialog.jsx
brmscheiner: yeah, this work could definitely be helpful for building some other text-heavy pages, plus this file is already super long. i think we should move these component definitions to the `src/components` directory. also i don't think we should name components the same thing as html elements, i feel like that could create confusion. maybe Header1 or something?

src/components/report/TermsAndConditionsDialog.jsx
brmscheiner: yeah unless it's absolutely necessary to use makeStyles let's stick to one method everywhere (which for now is inline styles).

src/pages/reportSighting/ReportForm.jsx
brmscheiner: yep, although it's just for cleanliness so it's a toss up in a situation like this when the variable names are shorter and there's only one expression

brmscheiner (approved):
approved with minor changes

src/components/report/TermsAndConditionsDialog.jsx
Emily-Ke: Corrected in https://github.com/WildMeOrg/codex-frontend/pull/275/commits/3200d7f7d9098fd2d0fc758d0577d879396d72b9.

src/components/report/TermsAndConditionsDialog.jsx
Emily-Ke: Changed the names (https://github.com/WildMeOrg/codex-frontend/pull/275/commits/b1ed9bffd8c1f4ca8dab552d0b1c3d6cc81d9bac) and moved to a separate file (https://github.com/WildMeOrg/codex-frontend/pull/275/commits/1df0aabab58cae57d74e9ac3603a465a54de2011).

src/pages/reportSighting/ReportForm.jsx
Emily-Ke: Applied in https://github.com/WildMeOrg/codex-frontend/pull/275/commits/1bbeec8610cd7b331d896cfcc9525699518c4585. I changed the variable name though, so please make sure it matches the conventions.

src/components/TextContainers.jsx
```diff
@@ -0,0 +1,60 @@
...
+
+export function Paragraph({ children, ...rest }) {
+  return (
+    <Typography paragraph {...rest}>
+      {children}
+    </Typography>
+  );
+}
+
+export function ListItem({ children, ...rest }) {
```
Atticus29, line 50: I think ListItem is a material UI component we already use in quite a few places. I'd support a re-name of this one... dunno what I'd change it to, though. Litem lmao?

Atticus29 (commented):
Looks even better! Left one comment.

PR #276, opened 2022-03-10
Send sex null instead of empty string, standardize in one list

brmscheiner:


Atticus29 (approved):
LGTM. I think that we should make sure that the distinction between unknown and unset is clear to the end user somehow.

Emily-Ke (approved):
The UNSET behavior in the UI is a little strange. It shows up in the menu as an option, but when I select it, the dropdown looks as if it is empty. UNSET and empty are the same, but that feedback is slightly irregular.

brmscheiner:
@Emily-Ke good call I updated it to look like this to blank the input. 
image

PR #277, opened 2022-03-11
Individual merge

brmscheiner:
Loading individual data:
image

Individual data ready:
image

After mvp we will close in on our initial design, supporting more merge conflict fields and merging more than 2 individuals 
image


src/models/assetGroupSighting/useGetAGS.js
```diff
@@ -1,13 +1,13 @@
...
 
-export default function useGetAGS(assetGroupSightingId) {
+export default function useGetAGS(assetGroupSightingId, disabled) {
   return useGet({
     queryKey: getAGSQueryKey(assetGroupSightingId),
     url: `/asset_groups/sighting/as_sighting/${assetGroupSightingId}`,
     queryOptions: {
       refetchInterval: 5000,
-      enabled: Boolean(assetGroupSightingId),
+      enabled: Boolean(assetGroupSightingId) && !disabled,
```
brmscheiner, line 10: @hwindsor this will prevent those extra requests you were seeing

src/pages/merge/IndividualCard.jsx
```diff
@@ -0,0 +1,75 @@
...
+            padding: '4px 0 0 16px',
+          }}
+        >
+          {data ? (
+            <Text style={{ fontWeight: 'bold' }}>{defaultName}</Text>
+          ) : (
+            <Skeleton variant="text" height={30} width={140} />
+          )}
+          {data ? (
+            <Text variant="body2">{adoptionName || 'Steve'}</Text>
```
Atticus29, line 56: Maybe change Steve to something unnamed? And/or just remove that part entirely, since adoptionName will be null if it's missing anyway?

src/pages/merge/MergeIndividuals.jsx
```diff
@@ -0,0 +1,149 @@
...
+import IndividualCard from './IndividualCard';
+
+export default function MergeIndividuals() {
+  const { search } = useLocation();
+  useDocumentTitle('MERGE_INDIVIDUALS');
+
+  const searchParams = new URLSearchParams(search);
+  const individualIds = searchParams.getAll('i') || [];
+
+  const { data: mergeConflicts } = useGetMergeConflicts(
```
Atticus29, line 28: Might not be a bad idea to handle errors and loading associated with getMargeConflicts, useIndividual (1 & 2)?

src/pages/merge/MergeIndividuals.jsx
```diff
@@ -0,0 +1,149 @@
...
+  );
+  const { data: individual1Data } = useIndividual(
+    get(individualIds, '0', null),
+  );
+  const { data: individual2Data } = useIndividual(
+    get(individualIds, '1', null),
+  );
+  const individualData = [individual1Data, individual2Data];
+
+  const { mutate: mergeIndividuals, loading } = useMergeIndividuals();
```
Atticus29, line 39: same here with error handling?

src/pages/merge/IndividualCard.jsx
```diff
@@ -0,0 +1,75 @@
...
+          ) : (
+            <Skeleton variant="text" height={20} width={80} />
+          )}
+          {data && showSex && (
+            <Text style={{ marginTop: 16 }}>
+              <Text variant="body2" component="span">{`Sex: `}</Text>
```
Atticus29, line 62: en.jsonify

src/pages/merge/MergeIndividuals.jsx
```diff
@@ -0,0 +1,149 @@
...
+                individualData={individualData}
+              />
+            )}
+          </Grid>
+        )}
+        <Grid item style={{ marginTop: 16 }}>
+          <Button
+            display="primary"
+            id="MERGE_INDIVIDUALS"
+            loading={loading}
```
Atticus29, line 128: loading should likely be a composite of all of the loadings?

src/hooks/useGet.js
```diff
@@ -22,6 +22,7 @@ export default function useGet({
   url,
   data,
   params,
+  method = 'get',
```
Emily-Ke, line 25: I don't love a function called `useGet` that is not making a GET request.

src/pages/merge/ResolutionSelector.jsx
```diff
@@ -0,0 +1,87 @@
...
+  sex: {
+    labelId: 'SEX',
+    getProperty: individualData => individualData?.sex,
+    deriveChoices: sexes =>
+      validSexOptions.filter(o => sexes.includes(o.value)),
+  },
+  defaultName: {
+    labelId: 'NAME',
+    getProperty: individualData => {
+      const names = get(individualData, 'names', []);
```
Atticus29, line 24: This logic seems to get repeated a lot. A way to DRY?

src/pages/merge/ResolutionSelector.jsx
```diff
@@ -0,0 +1,87 @@
...
+        label: name,
+        value: name,
+      })),
+  },
+  nickname: {
+    labelId: 'ALIAS', // untested!
+    getProperty: individualData => {
+      const names = get(individualData, 'names', []);
+      const defaultNameObject = names.find(
+        name => name?.context === 'Nickname',
```
Emily-Ke, line 41: I think 'Nickname' is supposed to be lower case.

src/pages/merge/ResolutionSelector.jsx
```diff
@@ -0,0 +1,87 @@
...
+  const individualProperties = individualData.map(individual =>
+    selectorSchema.getProperty(individual),
+  );
+  const choices = selectorSchema.deriveChoices(individualProperties);
+
+  return (
+    <FormControl style={{ width: 320, marginTop: 12 }}>
+      <InputLabel>
+        <FormattedMessage id={selectorSchema.labelId} />
+        {' *'}
```
Atticus29, line 69: ?

src/pages/merge/utils/formDataUtils.js
```diff
@@ -0,0 +1,34 @@
...
+export function derivePropertyOverrides(
+  formData,
+  showSexInput,
+  showDefaultNameInput,
+  showNicknameInput,
+) {
+  const overrides = {};
+  if (showSexInput) overrides.sex = formData?.sex;
+  if (showDefaultNameInput) {
+    set(
```
Atticus29, line 12: TIL lodash set. Cool.

src/pages/merge/MergeIndividuals.jsx
```diff
@@ -0,0 +1,149 @@
...
+              const propertyOverrides = derivePropertyOverrides(
+                formData,
+                showSexInput,
+                showDefaultNameInput,
+                showNicknameInput,
+              );
+
+              mergeIndividuals({
+                targetIndividualGuid: individual1Data?.guid,
+                fromIndividualGuids: [individual2Data?.guid],
```
Emily-Ke, line 140: What do you think about using these types of identifiers for the local variables?
For example, `targetIndividualData` instead of `individual1Data`.

Emily-Ke (commented):
1. Do you want to do a pass to rename `nickname` and (maybe) `defaultName` variables to `adoptionName` and `firstName`? It looks like you used `adoptionName` in one place already.
2. The `IndividualCards` are squashing the images for me. I changed the flex on the top one so that you can see the normal resolution. The bottom one is how they are displaying naturally. I think that you want to crop the image, but not squash it.
merge01

Atticus29 (changes requested):
Really minor stuff, and several questions.

src/hooks/useGet.js
brmscheiner: cc @Atticus29 yeah it's the worst but elasticsearch requests for data function exactly like GET requests but for weird reasons use the POST method instead. open to other suggestions but this felt clean and not-toooo-dirty to me

src/models/individual/useGetMergeConflicts.js
```diff
@@ -0,0 +1,14 @@
+import { getMergeConflictsQueryKey } from '../../constants/queryKeys';
+import useGet from '../../hooks/useGet';
+
+export default function useGetMergeConflicts(individualIds) {
+  return useGet({
+    queryKey: getMergeConflictsQueryKey(individualIds),
+    method: 'post',
```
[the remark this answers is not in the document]
brmscheiner: yeah this is a really weird aspect of the API imho, since this post doesn't mutate data i dont want to treat it like we treat other post requests

src/models/individual/useMergeIndividuals.js
```diff
@@ -0,0 +1,24 @@
...
+    deriveUrl: ({ targetIndividualGuid }) =>
+      `/individuals/${targetIndividualGuid}/merge`,
+    deriveData: ({ fromIndividualGuids, propertyOverrides }) => ({
+      fromIndividualIds: fromIndividualGuids,
+      parameters: {
+        override: propertyOverrides,
+      },
+    }),
+    onSuccess: response => {
+      const mergedIndividualGuid = response?.data?.targetId;
```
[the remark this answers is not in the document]
brmscheiner: it's the target individuals' guid

src/pages/individual/Individual.jsx
```diff
@@ -160,19 +167,14 @@ export default function Individual() {
...
                 {
-                  id: 'edit-profile',
-                  onClick: () => setEditingProfile(true),
-                  label: 'Edit profile',
-                },
-                {
-                  id: 'view-history',
-                  onClick: Function.prototype,
-                  label: 'View history',
+                  id: 'merge',
```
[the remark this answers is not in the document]
brmscheiner: because it doesn't do anything! hehe i've been cutting down on buttons that dont do anything as we approach release

src/pages/individual/MergeDialog.jsx
```diff
@@ -0,0 +1,48 @@
...
+          selectedIndividualId={selectedIndividualId}
+          setSelectedIndividualId={setSelectedIndividualId}
+        />
+      </DialogContent>
+      <DialogActions>
+        <Button display="basic" onClick={onCloseDialog} id="CLOSE" />
+        <ButtonLink
+          display="primary"
+          disabled={!selectedIndividualId}
+          href={`/merge?i=${individualGuid}&i=${selectedIndividualId}`}
```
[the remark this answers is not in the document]
brmscheiner: yeah, these are URL search params instead of `/:variables`, i set it up this way so we could support an unlimited number of individuals being merged

src/pages/merge/IndividualCard.jsx
```diff
@@ -0,0 +1,75 @@
...
+  const safeNames = data?.names || [];
+  const defaultName =
+    deriveName(safeNames, 'defaultName') || 'Unnamed individual';
+  const adoptionName = deriveName(safeNames, 'nickname');
+
+  const showSex = mergeConflicts?.sex;
+  const sexObject = sexOptions.find(
+    option => option.value === data?.sex,
+  );
+  const sexLabelId = sexObject?.mergeLabelId || sexObject?.labelId;
```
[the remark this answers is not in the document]
brmscheiner: In the dropdown the option to set sex to null appears as an empty string, in the merge interface it will show up as "Value not set"

src/pages/merge/IndividualCard.jsx
brmscheiner: oh yeah this was for testing lololol

src/pages/merge/IndividualCard.jsx
```diff
@@ -0,0 +1,75 @@
...
+          }}
+        >
+          {data ? (
+            <Text style={{ fontWeight: 'bold' }}>{defaultName}</Text>
+          ) : (
+            <Skeleton variant="text" height={30} width={140} />
+          )}
+          {data ? (
+            <Text variant="body2">{adoptionName || 'Steve'}</Text>
+          ) : (
```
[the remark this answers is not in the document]
brmscheiner: just nothing

src/pages/merge/MergeIndividuals.jsx
brmscheiner: probably not here, since the button action will only be in a loading state when a merge is taking place

src/pages/merge/MergeIndividuals.jsx
brmscheiner: i can do that

src/pages/merge/ResolutionSelector.jsx
brmscheiner: asterisk to designate they are required

src/pages/merge/ResolutionSelector.jsx
brmscheiner: nice catch

src/utils/nameUtils.js
```diff
@@ -0,0 +1,9 @@
+export function deriveIndividualName(
+  individualData,
+  context = 'defaultName',
+  fallback = undefined,
```
Emily-Ke, line 4: `fallback` will default to undefined automatically. It shouldn't require an explicit assignment.

src/utils/nameUtils.js
```diff
@@ -0,0 +1,9 @@
...
+) {
+  const names = individualData?.names || [];
+  const nameObject = names.find(n => n?.context === context) || {};
+  return nameObject?.value || fallback;
```
Emily-Ke, line 8: What is the scenario where `nameObject` is nullish?

brmscheiner: since the individual data comes back from the API we have to assume it could have pretty much anything wrong with it. for example if we received 

```
{
  ...individualProperties...
  name: [{ name1... }, undefined, { name2... }
}
```

we will hopefully render something janky around the weird value rather than fully crashing the page.

brmscheiner:
@Emily-Ke fixed the image stretching. note even this is temporary: eventually almost all photos containing animals will have zoom/pan controls, and in this case maybe some buttons to scroll through other photos of the same individual. (it's not so much for novelty, sometimes being able to look closely at the pattern on an animal can be an important part of the decision making process)

not gonna change to adoptionName/firstName right now. gonna do it everywhere in the app while testing against the migration data in an upcoming PR

src/utils/nameUtils.js
Emily-Ke: What you currently have in line 7 (`const nameObject = names.find(n => n?.context === context) || {};`) should be handling that though. Either `names.find` will find the correct name object, in which case `nameObject` is an object with at least a context field, or `names.find` will return `undefined`, in which case your or operator sets `nameObject` to an empty object. I'm not seeing a case when `nameObject` is not an object.

src/hooks/useGet.js
Emily-Ke: If this is only going to be used for `useGetMergeConflicts`, then my preference would be to just duplicate the logic in that function.

If this is required for all elasticsearch interaction, then maybe create a new function `useElasticSearchGet` and either duplicate the logic or abstract and share it.

If you prefer to leave it as is, I prefer renaming the function to remove the word 'Get'.

src/utils/nameUtils.js
brmscheiner: fair enough. do you think `?.` hurts anything?

src/hooks/useGet.js
brmscheiner: okay it is switched to `useFetch` everywhere. was this worth the time? probably not. but it's done now.

Atticus29 (approved):
Way to make "fetch" happen.

Emily-Ke (approved):

PR #278, opened 2022-03-12
Dex 663 qa feedback and DEX-808

Atticus29:
Responds to all of JH's QA feedback from https://wildme.atlassian.net/browse/DEX-663:

> In the Notifications modal, clicking “Mark as Read” can cause notifications to disappear or change their button to a grayed out DONE.

Per discussions with Ben, all notifications have a view button, but some of these view buttons navigate the user to their home page, while others pull up a dialog with actions.

> "Mark as Read” button can change size depending on length of text next to it:

In the process of troubleshooting this, I kind of accidentally added a <div> around the button. This coincidentally seems to force a standardization of button size. If someone who understands css better than me know why this happened, I would be eager to know that.

> In “View All Notifications”, clicking on an entry does not then turn it gray to show it has been read. Seems to be persisting on the backend correctly though as a refresh then shows the read entry as gray. 

The refreshNotifications() method originally being called was something that we recently deprecated from the useNotification (and other useGet-based) hooks. It now actually refreshes by invalidating query keys.

> User Manager-revoked collaboration falsely tells both affected users that the other user hates them and revoked the collaborations. Should say that the collaboration was revoked by the User Manager like it does when the collab is created by the user manager.

This is a back end bug. I created a ticket for them, and a separate FE ticket for me to follow up on, thereby decoupling this issue from the current ticket.

Note that I also went ahead and tackled https://wildme.atlassian.net/browse/DEX-808 in this PR.
The issue here was that the little popover window or whatever it's called was using the xAnchor point both as its `open` boolean and as its anchor point. When someone dismissed it, it was setting the xAnchor point to null, which I guess was generously being interpreted as 0 on the x-axis. I resolved this by decoupling the xAnchor from the thing that controls `open`.

### DIFF
--- a/src/components/AuthenticatedAppHeader/NotificationsPane.jsx
+++ b/src/components/AuthenticatedAppHeader/NotificationsPane.jsx
@@ -1,6 +1,8 @@
 import React, { useState } from 'react';
 import { useIntl } from 'react-intl';
+import { useHistory } from 'react-router-dom';
 import { get } from 'lodash-es';
+import { useQueryClient } from 'react-query';
 
 import Grid from '@material-ui/core/Grid';
 import Popover from '@material-ui/core/Popover';
@@ -17,6 +19,7 @@ import shane from '../../assets/shane.jpg';
 import { notificationSchema } from '../../constants/notificationSchema';
 import { calculatePrettyTimeElapsedSince } from '../../utils/formatters';
 import { notificationTypes } from '../dialogs/notificationDialogUtils';
+import queryKeys from '../../constants/queryKeys';
 
 function renderDisplayTextBasedOnMessageType(
   currentNotificationSchema,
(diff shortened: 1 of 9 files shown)


src/components/AuthenticatedAppHeader/NotificationsPane.jsx
```diff
@@ -35,162 +38,161 @@ function renderDisplayTextBasedOnMessageType(
...
-    <Popover
-      open={Boolean(anchorEl)}
-      anchorOrigin={{ vertical: 'bottom', horizontal: 'right' }}
-      PaperProps={{ style: { marginTop: -8 } }}
-      transformOrigin={{ horizontal: 'right', vertical: 'top' }}
-      anchorEl={anchorEl}
-      onClose={() => setAnchorEl(null)}
-    >
+    <div>
       {Boolean(activeCollaborationNotification) && (
```
Atticus29, line 59: Moved this out of the popover so that the setShouldOpen(false) therein wouldn't also dismiss the dialog.

src/constants/notificationSchema.js
```diff
@@ -70,30 +72,33 @@ notificationSchemaPlaceholder[
...
 };
 notificationSchemaPlaceholder[
   notificationTypeNames.individual_merge_complete
 ] = {
   titleId: 'INDIVIDUAL_MERGE_COMPLETE_TITLE',
   notificationMessage: 'INDIVIDUALS_MERGE_COMPLETE_MESSAGE',
   moreDetailedDescription: 'INDIVIDUALS_MERGE_COMPLETE_MESSAGE',
-  viewInNotificationPane: false,
+  showNotificationDialog: false,
+  buttonPath: '/individuals/', // TODO figure out how to pass individual ID to this
```
Atticus29, line 102: These TODOs having to do with individual notifications will be fleshed out once we can use the UI to merge individuals.

src/components/dialogs/NotificationDetailsDialog.jsx
```diff
@@ -26,14 +26,18 @@ export default function NotificationDetailsDialog({
...
+    ['message_values', 'user1_name'],
+    'Unnamed User',
+  );
+  if (user1Name === '') user1Name = 'Unnamed User';
+  let user2Name = get(
+    notification,
+    ['message_values', 'user2_name'],
+    'Unnamed User',
+  );
+  if (user2Name === '') user2Name = 'Unnamed User';
```
brmscheiner, line 40: you could do this instead (since '' is falsey):
```const user2Name = notification?.message_values?.user2_name || 'Unnamed User';```

File: src/constants/notificationSchema.js (hunk 2 above)
Comment (brmscheiner) [reply src/constants/notificationSchema.js:102]: yep all the merge notifications will need to get worked on once you can actually merge. this is fine for now :)

File: src/pages/notifications/Notifications.jsx [hunk 4]
```diff
@@ -89,14 +91,16 @@ export default function Notifications() {
...
+                const user1Name = get(
+                  notification,
+                  ['message_values', 'user1_name'],
+                  'Unnamed User',
+                );
+                const user2Name = get(
+                  notification,
+                  ['message_values', 'user2_name'],
+                  'Unnamed User',
+                );
```
brmscheiner, line 103: this is the same logic as above. maybe a utility function would be good?

brmscheiner (approved):
small changes, looks good overall

PR #279, opened 2022-03-14
DEX-811 missing "report sighting success" page

Emily-Ke:
- After successfully POSTing a sighting, an unauthenticated user is redirected to `/report/success` rather than to the pending sighting page that they are not authorized to see.
- Because only unauthorized users see the Success page, I removed it from the AuthenticatedSwitch, and removed the authorization conditions within the Success component.


brmscheiner (approved):

Atticus29 (approved):
LGMT

PR #280, opened 2022-03-16
Flatfile hotfix

brmscheiner:


Atticus29 (approved):
Weird to me that this fixes anything. What was broken before?

src/pages/bulkImport/utils/useBulkImportFields.js
```diff
@@ -32,6 +32,13 @@ const utcOffsetValidator = {
   error: 'Not a valid timezone (UTC offset format)',
 };
 
+function deriveLabel(field, intl) {
+  const label = field?.labelId
+    ? intl.formatMessage({ id: field.labelId })
+    : field.label;
```
Emily-Ke, line 38: If `field?.labelId` can be null because field can be null, then this could still error.

brmscheiner: thanks. that wasn't the original issue but still worth fixing.

brmscheiner:
@Atticus29 there is a bad custom field creating by backend tests that is hanging out on the QA server. it has no schema and therefore no label. this attempts to make the bulk import UI continue to work in such cases

src/pages/bulkImport/utils/useBulkImportFields.js
brmscheiner: @Emily-Ke pushed

Atticus29: As, in the bottom part should be:

`field?.label `
as well?

Or does:
`null ? something: somethingElse `
inherently cause an error?

brmscheiner: the first one! `?.`

Atticus29: I'm seeing from the new commit that it's the former.

Emily-Ke (approved):

PR #281, opened 2022-03-18
DEX-847 remain on tab after committing AGS

Emily-Ke:
- Adds the current hash to the redirect URL after a successful AGS commit so that the user stays on the same tab they were on when they committed the AGS.

### DIFF
--- a/src/pages/sighting/AssetGroupSighting.jsx
+++ b/src/pages/sighting/AssetGroupSighting.jsx
@@ -1,5 +1,5 @@
 import React, { useEffect } from 'react';
-import { useParams, useHistory } from 'react-router-dom';
+import { useParams, useHistory, useLocation } from 'react-router-dom';
 
 import SightingCore from './SightingCore';
 import useGetAGS from '../../models/assetGroupSighting/useGetAGS';
@@ -7,6 +7,7 @@ import useGetAGS from '../../models/assetGroupSighting/useGetAGS';
 export default function AssetGroupSighting() {
   const history = useHistory();
   const { id } = useParams();
+  const { hash } = useLocation();
 
   const {
     data: assetGroupSightingData,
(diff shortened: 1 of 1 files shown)


src/pages/sighting/AssetGroupSighting.jsx
```diff
@@ -1,12 +1,13 @@
...
-import { useParams, useHistory } from 'react-router-dom';
+import { useParams, useHistory, useLocation } from 'react-router-dom';
 
 import SightingCore from './SightingCore';
 import useGetAGS from '../../models/assetGroupSighting/useGetAGS';
 
 export default function AssetGroupSighting() {
   const history = useHistory();
   const { id } = useParams();
+  const { hash } = useLocation();
```
brmscheiner, line 10: if there is no hash, does this give empty string? if so feel free to merge. otherwise i think you might get a weird URL like `/sightings/<guid>undefined`

brmscheiner (approved):

src/pages/sighting/AssetGroupSighting.jsx
Emily-Ke: Yes. When I click "Commit" from `/pending-sightings/facd3479-88dc-4e3d-91f0-214e37e0b8a2`, I am redirected to `/sightings/4d38de61-ea9d-4ff8-a088-7c96fa5117d2`

Do you think that we need a check anyway to make sure that if there is no hash, it is an empty string?

Atticus29 (approved):
LGTM other than Ben's comment

src/pages/sighting/AssetGroupSighting.jsx
brmscheiner: nah, if that's how react-router works that's how it works! i was just too lazy to run it on my system and check :)

PR #282, opened 2022-03-18
Dex 763 relationships in sight settings rebased

Atticus29:
Phew!

Adds a relationship editor to the "manage fields → Configurable default fields → Relationship".

Accommodates updating values, deleting values, and adding values for both relationship "categories" (back end calls them "types", I think) (e.g., "familial") and for "relationships" (don't remember what the back end calls them) (e.g., "mother").

Note that there is quite a bit of code that I imagine could both be dried up and better named, so please be on the lookout for areas that are hard to read; I'd like to improve those.


src/pages/fieldManagement/settings/defaultFieldComponents/Editors.jsx
```diff
@@ -1,42 +1,62 @@
...
-//   onSubmit,
-//   formSettings,
-//   setFormSettings,
-//   ...rest
-// }) {
-//   const speciesOptions = get(formSettings, 'relationships', []);
+  const transformedRelationshipOptions = map(
+    relationshipOptions,
+    (relationshipArr, key) => {
+      const currentObj = {};
```
Atticus29, line 23: I doubt that this was necessary in retrospect. But my brain worked much better when the categories/types were tranformed in this way.

src/pages/fieldManagement/settings/defaultFieldComponents/Editors.jsx
```diff
@@ -1,42 +1,62 @@
...
+    >
+      <RelationshipEditor
+        schema={{ labelId: 'RELATIONSHIPS' }}
+        value={transformedRelationshipOptions}
+        onChange={relationships => {
+          const formSettingObj = {};
+          const reformattedRelationships = relationships.reduce(
+            (memo, obj) => {
+              let currentObj = {};
+              currentObj[(obj?.category)] = obj?.relationships;
```
Atticus29, line 46: Just undoing the thing from above on the way back out, as it were.

src/pages/fieldManagement/settings/defaultFieldComponents/RelationshipEditor.jsx
```diff
@@ -0,0 +1,304 @@
...
+  const nonTargeCategoryObjects = filter(
+    currentAllRelationshipObject,
+    currentRelationshipCategory =>
+      currentRelationshipCategory.category !==
+      modifiedTargetCategoryObj?.category,
+  );
+  return [...nonTargeCategoryObjects, modifiedTargetCategoryObj];
+}
+
+function deleteCategory(allRelationshipsObj, category) {
```
Atticus29, line 29: In keeping the existing pattern elsewhere, all of these delete and edit methods return the full object containing all relationship information.

src/pages/fieldManagement/settings/defaultFieldComponents/RelationshipEditor.jsx
```diff
@@ -0,0 +1,304 @@
...
+  allRelationshipsObj,
+  currentCategory,
+  relationship,
+) {
+  const targetRelationshipCategoryObjArr = filter(
+    allRelationshipsObj,
+    currentRelationshipCategory =>
+      currentRelationshipCategory.category === currentCategory,
+  );
+  // if (targetRelationshipCategoryObjArr?.length > 1) return; // TODO how do we handle duplicate relationship categories/types?
```
Atticus29, line 46: Post MVP? Just treat them all as independent, even if they're the same exact? Prevent users from creating categories or relationships that already exist?

src/pages/fieldManagement/settings/defaultFieldComponents/RelationshipEditor.jsx
```diff
@@ -0,0 +1,304 @@
...
+          endAdornment: (
+            <InputAdornment position="end">
+              <IconButton
+                size="small"
+                onClick={() => {
+                  onChange(
+                    addRelationshipToCategory(
+                      allRelationshipsObj,
+                      category,
+                      '',
```
Atticus29, line 230: We could easily make this something like, "Enter new relationship name here"

src/pages/fieldManagement/settings/DefaultFieldTable.jsx
```diff
@@ -115,22 +115,30 @@ export default function DefaultFieldTable({
...
                 formSettings.species,
               ).then(success => {
                 if (success) onCloseEditor();
               });
             }
+            if (editField?.id === 'relationship') {
+              putSiteSetting(
+                editField.backendPath,
+                formSettings.relationships,
+              ).then(success => {
```
brmscheiner, line 138: let's stick to `async/await`. one asynchronous api is enough!

brmscheiner, line 138: do you mind changing these to `async/await`? i want to use the same asynchronous patterns everywhere and i dunno why i deviated here.

brmscheiner: also it does seem like we could DRY this whole block up a little...

```
const relevantFormSettingsProperty = ... // you would set this part up and give it a better name 
const requestSuccessful = await putSiteSetting(
    editField.backendPath,
    relevantFormSettingsProperty,
);
if (requestSuccessful) onCloseEditor();
```

src/pages/fieldManagement/settings/defaultFieldComponents/Editors.jsx
```diff
@@ -1,42 +1,62 @@
...
-//         });
-//       }}
-//       {...rest}
-//     />
-//   );
-// }
+  return (
+    <ConfigureDefaultField
+      onClose={onClose}
+      onSubmit={() => {
```
brmscheiner, line 33: `onSubmit={onSubmit}`

src/pages/fieldManagement/settings/defaultFieldComponents/Editors.jsx
brmscheiner: I'd really rather not do this if it's not necessary :D

src/pages/fieldManagement/settings/defaultFieldComponents/RelationshipEditor.jsx
```diff
@@ -0,0 +1,304 @@
...
+import InputAdornment from '@material-ui/core/InputAdornment';
+import NewChildIcon from '@material-ui/icons/AddCircle';
+import TextInput from '../../../../components/inputs/TextInput';
+import DeleteButton from '../../../../components/DeleteButton';
+import Button from '../../../../components/Button';
+import Text from '../../../../components/Text';
+
+const intl = useIntl();
+
+function recomposeAllRelationshipObjectFromModifiedCategoryObject(
```
brmscheiner, line 16: i know names like this are very clear but at some point they are just too long lmao

brmscheiner: Supercalifragilisticexpialidocious is 34 characters and hereafter the upper limit on variable names. yours is 56!

src/pages/fieldManagement/settings/defaultFieldComponents/RelationshipEditor.jsx
```diff
@@ -0,0 +1,304 @@
...
+  currentAllRelationshipObject,
+  modifiedTargetCategoryObj,
+) {
+  const nonTargeCategoryObjects = filter(
```
brmscheiner, line 20: typo

src/pages/fieldManagement/settings/defaultFieldComponents/RelationshipEditor.jsx
```diff
@@ -0,0 +1,304 @@
...
+      singleRelationshipObj?.category !== category,
+  );
+}
+
+function deleteRelationship(
+  allRelationshipsObj,
+  currentCategory,
+  relationship,
+) {
+  const targetRelationshipCategoryObjArr = filter(
```
brmscheiner, line 41: there are certainly exceptions, but let's try to cut down on including JS types as variable names (object and array here). for example i think this one could just be called `targetRelationshipCategories`

src/pages/fieldManagement/settings/defaultFieldComponents/RelationshipEditor.jsx
brmscheiner: well since the categories are keys for an object it shouldn't be possible to duplicate them. but we should make sure things aren't buggy if they try. 

different story for the relationships since they are just an array. i do think they should be prohibited from duplicating them.

brmscheiner (commented):
sorry to stop reviewing halfway through but i need to hit the hay. i'll finish reviewing soon!

src/pages/fieldManagement/settings/defaultFieldComponents/RelationshipEditor.jsx
```diff
@@ -0,0 +1,304 @@
...
+import { FormattedMessage } from 'react-intl';
+import IconButton from '@material-ui/core/IconButton';
+import InputAdornment from '@material-ui/core/InputAdornment';
+import NewChildIcon from '@material-ui/icons/AddCircle';
+import TextInput from '../../../../components/inputs/TextInput';
+import DeleteButton from '../../../../components/DeleteButton';
+import Button from '../../../../components/Button';
+import Text from '../../../../components/Text';
+
+const intl = useIntl();
```
Emily-Ke, line 14: For me, this line is preventing the hot updater from working.
```
Invalid hook call. Hooks can only be called inside of the body of a function component.
```

src/pages/fieldManagement/settings/defaultFieldComponents/RelationshipEditor.jsx
```diff
@@ -0,0 +1,304 @@
...
+  singleRelationshipObj,
+  onChange,
+) {
+  const category = singleRelationshipObj?.category;
+  const relationships = singleRelationshipObj?.relationships;
+  return (
+    <div style={{ marginTop: 10 }}>
+      <TextInput
+        width={240}
+        schema={category}
```
Emily-Ke, line 208: `category` is just a string that is a user input. Why pass it as the `schema`?

src/pages/fieldManagement/settings/defaultFieldComponents/RelationshipEditor.jsx
```diff
@@ -0,0 +1,304 @@
...
+  targetRelationshipCategoryObj.relationships = currentRelationships;
+  return recomposeAllRelationshipObjectFromModifiedCategoryObject(
+    allRelationshipsObj,
+    targetRelationshipCategoryObj,
+  );
+}
+
+function createCategory(allRelationshipsObj) {
+  const newCategoryObj = {
+    category: intl.formatMessage({ id: 'ENTER_NEW_CATEGORY_NAME' }),
```
Emily-Ke, line 96: I would prefer to not have to delete this text. When adding a region, we just add an empty text field. If you want to make it clear that this level of text field is a category and not a relationship, I think that a placeholder would be better than setting the actual value.

src/pages/fieldManagement/settings/defaultFieldComponents/RelationshipEditor.jsx
Emily-Ke: I would rather leave it blank and add a placeholder if you feel that it needs it.

src/pages/fieldManagement/settings/defaultFieldComponents/Editors.jsx
Atticus29: Whoops... a vestige from a print statement that got deleted.

src/pages/fieldManagement/settings/defaultFieldComponents/RelationshipEditor.jsx
brmscheiner: ahh shoot, this comment made me realize we are using the old form inputs here. everything in `src/components/inputs` is deprecated in favor of `src/components/fields/edit`. these components take a schema prop because they are set up to render form elements based on houston-supplied schemas (both the new and old version)

however, as seen here, it's possible to easily render a form element by "faking" a schema using only the schema properties necessary to get it looking the way you want. it's not clean and i'm not sure how i feel about it in the long haul but i think it's ok for mvp. 

better alternative is probably to use the `TextField` component straight from material-ui

brmscheiner: at some point i really need to migrate over the remaining things that are using `src/components/inputs` and delete this folder (there aren't many left but i never had time to get 'em all)

src/pages/fieldManagement/settings/defaultFieldComponents/Editors.jsx
Atticus29: Done

src/pages/fieldManagement/settings/defaultFieldComponents/RelationshipEditor.jsx
Atticus29: Done. Handled the relationship duplication in the actual button submission since it creates undesired errors while the user is making progress on typing the new potential relationship.

src/pages/fieldManagement/settings/defaultFieldComponents/RelationshipEditor.jsx
Atticus29: Done

src/pages/fieldManagement/settings/DefaultFieldTable.jsx
Atticus29: Actually, this ended up being where I handled relationship deduplication. So, it's not as repetitive as it was before.

src/pages/fieldManagement/settings/defaultFieldComponents/Editors.jsx
Atticus29: Done

src/pages/fieldManagement/settings/defaultFieldComponents/RelationshipEditor.jsx
Atticus29: Done.

src/pages/fieldManagement/settings/defaultFieldComponents/RelationshipEditor.jsx
Atticus29: Done... with the exception of there needing to be a distinction between renderSingleRelationshipObj and renderSingleRelationship

src/pages/fieldManagement/settings/defaultFieldComponents/RelationshipEditor.jsx
Atticus29: Done

src/pages/fieldManagement/settings/defaultFieldComponents/RelationshipEditor.jsx
```diff
@@ -0,0 +1,304 @@
...
+  );
+}
+
+function renderSingleRelationshipObj(
+  allRelationshipsObj,
+  singleRelationshipObj,
+  onChange,
+) {
+  const category = singleRelationshipObj?.category;
+  const relationships = singleRelationshipObj?.relationships;
```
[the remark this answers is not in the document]
Atticus29: N/A in new refactor.

src/pages/fieldManagement/settings/defaultFieldComponents/RelationshipEditor.jsx
```diff
@@ -0,0 +1,304 @@
...
+                    deleteCategory(allRelationshipsObj, category),
+                  );
+                }}
+              />
+            </InputAdornment>
+          ),
+        }}
+      />
+      {relationships.map(relationship =>
+        renderSingleRelationship(
```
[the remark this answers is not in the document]
Atticus29: Added key to the div. Warning seems to have disappeared.

src/pages/fieldManagement/settings/defaultFieldComponents/RelationshipEditor.jsx
Atticus29: Keeping as is for now. This will change in a future refactor.

src/pages/fieldManagement/settings/defaultFieldComponents/RelationshipEditor.jsx
Atticus29: Done

src/pages/fieldManagement/settings/DefaultFieldTable.jsx
```diff
@@ -114,30 +114,57 @@ export default function DefaultFieldTable({
...
+                  return { ...memo, ...newRelationshipChunk };
+                },
+                {},
+              );
+              let newSiteSettings = {};
+              newSiteSettings['relationships'] = { ...newSettings };
+              formSettings.relationships = newSiteSettings; // I know that this is bad practice, but just passing newSiteSettings below did not seem to do the trick, and I don't know why
+              const success = await putSiteSetting(
                 editField.backendPath,
                 formSettings.relationships,
```
Atticus29, line 165: this is where I tried to use newSiteSettings instead.

src/pages/fieldManagement/settings/DefaultFieldTable.jsx
```diff
@@ -114,22 +114,57 @@ export default function DefaultFieldTable({
...
+                      return (
+                        currentRelationships.indexOf(entry) === index
+                      );
+                    },
+                  );
+                  if (
+                    currentRelationships.length >
+                    dedupedRelationships.length
+                  ) {
+                    alert(
```
brmscheiner, line 147: dunno how i feel about using browser alerts... this is okay for now but eventually i want to think more deeply about this

src/pages/fieldManagement/settings/DefaultFieldTable.jsx
```diff
@@ -114,22 +114,57 @@ export default function DefaultFieldTable({
...
+                      intl.formatMessage({
+                        id:
+                          'DUPLICATE_RELATIONSHIPS_FOUND_AND_REMOVED',
+                      }),
+                    );
+                  }
+                  let newRelationshipChunk = {};
+                  newRelationshipChunk[key] = dedupedRelationships;
+                  return { ...memo, ...newRelationshipChunk };
```
brmscheiner, line 156:
```
return {
  ...memo,
  [key]: dedupedRelationships,
}
```

src/pages/fieldManagement/settings/DefaultFieldTable.jsx
brmscheiner: i'd like to figure this out before merge - no reason to leave it dirty like this. what happens if you do 

```
const newSiteSettings = { 
    relationships: { ...newSettings }
}
const success = await putSiteSetting(editField.backendPath, newSiteSettings);
```

src/pages/fieldManagement/settings/defaultFieldComponents/Editors.jsx
```diff
@@ -3,40 +3,38 @@ import { get } from 'lodash-es';
...
+  formSettings,
+  setFormSettings,
+  children,
+  ...rest
+}) {
+  const transformedRelationshipOptions = get(
+    formSettings,
+    'relationships',
+    formSettings,
+  );
```
brmscheiner, line 20: this seems weird. maybe just double-check?

src/pages/fieldManagement/settings/defaultFieldComponents/Editors.jsx
```diff
@@ -3,40 +3,38 @@ import { get } from 'lodash-es';
...
-// }
+  return (
+    <ConfigureDefaultField onClose={onClose} onSubmit={onSubmit} open>
+      <RelationshipEditor
+        schema={{ labelId: 'RELATIONSHIPS' }}
+        value={transformedRelationshipOptions}
+        onChange={relationships => {
+          const formSettingObj = {};
+          formSettingObj['relationships'] = relationships;
+          setFormSettings(formSettingObj);
```
brmscheiner, line 30: setFormSettings({ relationships })

src/pages/fieldManagement/settings/defaultFieldComponents/Editors.jsx
```diff
@@ -3,40 +3,38 @@ import { get } from 'lodash-es';
...
-
-// export function RelationshipEditor({
-//   onClose,
-//   onSubmit,
-//   formSettings,
-//   setFormSettings,
-//   ...rest
-// }) {
-//   const speciesOptions = get(formSettings, 'relationships', []);
+export function RelationshipEditorWrapper({
```
brmscheiner, line 8: so a future programmer could easily grab "RelationshipEditor" try to start working with it. might take someone a while to realize that they need to be using the wrapper and not the RelationshipEditor itself. for this reason i prefer something like 

```
RelationshipEditor
 |_ RelationshipEditorCore 
```

src/pages/fieldManagement/settings/defaultFieldComponents/RelationshipEditor.jsx
```diff
@@ -0,0 +1,235 @@
...
+function deleteRelationship(
+  allRelationships,
+  currentCategory,
+  relationship,
+) {
+  const targetRelationshipCategory = get(
+    allRelationships,
+    currentCategory,
+  );
+  const relationshipsWithoutTarget = targetRelationshipCategory.filter(
```
brmscheiner, line 27: so if `get` fails to find anything on line 23, then `null.filter()` will crash the app here. so either (1) you are confident that currentCategory will always exist on the object because we aren't depending on anything external and we can just do 
```
const targetRelationshipCategory = allRelationships[currentCategory];
```
or (2) we can't rely on these things and need to provide a better fallback
```
  const targetRelationshipCategory = get(
    allRelationships,
    currentCategory,
    []
  );
```

src/pages/fieldManagement/settings/defaultFieldComponents/RelationshipEditor.jsx
```diff
@@ -0,0 +1,235 @@
...
+    entry => entry !== relationship,
+  );
+  allRelationships[currentCategory] = relationshipsWithoutTarget;
+  return allRelationships;
```
brmscheiner, line 31: it's usually bad practice to return a mutated version of the same object in functions like this. it's one of the areas JS is weird and debugging can get really hard. best practice is something like 

```
return {
    ...allRelationships,
    [currentCategory]: relationshipsWithoutTarget
}
```

src/pages/fieldManagement/settings/defaultFieldComponents/RelationshipEditor.jsx
```diff
@@ -0,0 +1,235 @@
...
+}
+
+function addRelationshipToCategory(
+  allRelationships,
+  currentCategory,
+  newRelationship,
+) {
+  const currentRelationships = get(allRelationships, currentCategory);
+  if (!currentRelationships.includes(newRelationship))
+    currentRelationships.push(newRelationship);
```
brmscheiner, line 78: same thing with arrays, instead of pushing to arrays it's usually better to create a new array with the item added

src/pages/fieldManagement/settings/defaultFieldComponents/RelationshipEditor.jsx
```diff
@@ -0,0 +1,235 @@
...
+    </div>
+  );
+}
+
+function renderSingleRelationshipObj(
+  allRelationships,
+  targetKey,
+  onChange,
+) {
+  const relationships = get(allRelationships, targetKey);
```
brmscheiner, line 133: the map below will crash the app if `allRelationships[targetKey]` is not found. fallback to []

src/pages/fieldManagement/settings/defaultFieldComponents/RelationshipEditor.jsx
```diff
@@ -0,0 +1,235 @@
...
+  value,
+  onChange,
+  siteSettings, // unpacking to prevent passing to div
+  ...rest
+}) {
+  const justRelationshipsFormPart = get(
+    value,
+    'relationships',
+    value,
+  );
```
brmscheiner, line 197: this also seems weird to me. fallback to value if relationships.value isn't found?

src/pages/fieldManagement/settings/DefaultFieldTable.jsx
Atticus29: I figured that you would not love this. I initially tried using setError(), but the subsequent putSiteSetting() call ends up being successful and dismissing the dialog, so you see the "error" for a millisecond.
I also tried adding another hasDuplicate boolean before dismissal and ended up not using that either for a reason that I can't remember.
I guess we could do a nested dialog, but that seemed crazy. I'd like to chop that into a separate task if that's what is desired.

src/pages/fieldManagement/settings/DefaultFieldTable.jsx
Atticus29: Done. Good eye.

src/pages/fieldManagement/settings/defaultFieldComponents/Editors.jsx
Atticus29: The reason for this is that formSettings initially looks like {relationships:{}, species:{}, regions{}}, but in subsequent renderings is just the relationships subset. I think that it's totally ok that it loses species and region key-value pairs (do we call them elements or attributes for objects??), because the user cannot simultaneously edit all three anyway (i.e., the save button in the dialog ensures this). 
Given that the formSettings can enter that component one of two ways, this was how I solved that issue.
I tried dumber ways as well ⚰️ .

src/pages/fieldManagement/settings/defaultFieldComponents/Editors.jsx
Atticus29: This works because of 
```
const transformedRelationshipOptions = get(
    formSettings,
    'relationships',
    formSettings,
  );
```
, right? Otherwise, I don't understand why this worked.

src/pages/fieldManagement/settings/defaultFieldComponents/Editors.jsx
Atticus29: I like this idea a lot. Thanks. Done.

src/pages/fieldManagement/settings/defaultFieldComponents/RelationshipEditor.jsx
Atticus29: Good catch. Thanks. Done.

src/pages/fieldManagement/settings/defaultFieldComponents/RelationshipEditor.jsx
```diff
@@ -0,0 +1,235 @@
...
+    oldRelationshipName,
+  );
+  if (oldRelationshipIndex > -1)
+    currentRelationships[oldRelationshipIndex] = newRelationshipName;
+  allRelationships[category] = currentRelationships;
+  return allRelationships;
+}
+
+function createCategory(allRelationships) {
+  console.log(allRelationships);
```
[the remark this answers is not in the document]
Atticus29: Done. Thanks.

src/pages/fieldManagement/settings/defaultFieldComponents/RelationshipEditor.jsx
Atticus29: Certainly intentional. See above.

src/pages/fieldManagement/settings/defaultFieldComponents/RelationshipEditor.jsx
Atticus29: Thanks. Done.

src/pages/fieldManagement/settings/defaultFieldComponents/RelationshipEditor.jsx
Atticus29: Ok. What do you think of what I did here with the new commit?

src/pages/fieldManagement/settings/DefaultFieldTable.jsx
Atticus29: TIL that you can add a new key-value pair to an object without a string using [someVariable] as the key. Awesome! Very good to know this.

src/pages/fieldManagement/settings/defaultFieldComponents/RelationshipEditor.jsx
Atticus29: I always get super confused about what needs deep copies and what does not.

src/pages/fieldManagement/settings/DefaultFieldTable.jsx
Atticus29: Yeah... same issue that I was seeing earlier: it does not replace with the de-duplicated version! So weird!

src/pages/fieldManagement/settings/DefaultFieldTable.jsx
brmscheiner: let's merge this as-is now. but could you create a little ticket to look back at this file and assign it to me?

src/pages/fieldManagement/settings/DefaultFieldTable.jsx
brmscheiner: mmk, i will take a look at this at some point post-merge too

src/pages/fieldManagement/settings/defaultFieldComponents/Editors.jsx
brmscheiner: this seems really weird! is that how i had it set it up for species and regions as well? sounds like a mistake to me

Atticus29: I'm guessing that this has something to do with all of them belonging to the same form schema. IMHO they don't need to be.

brmscheiner (approved):

Emily-Ke (commented):
I expect that you plan to handle this, but please actually squash the `squash me` commits into something with a meaningful name before this is merged.

src/pages/fieldManagement/settings/DefaultFieldTable.jsx
Emily-Ke: In your onChange in defined in the RelationshipEditor, you are just setting the form state to relationships, which explains the behavior you are seeing. You need to include the entire form state. I think that all relationships are provided to the onChange, so the following should work:
```
onChange={relationships => {
    setFormSettings({ 
        ...formSettings,
        relationships
      });
  }}

```

src/pages/fieldManagement/settings/defaultFieldComponents/Editors.jsx
Emily-Ke: In your onChange in defined in the RelationshipEditor, you are just setting the form state to relationships, which explains the behavior you are seeing. You need to include the entire form state. I think that all relationships are provided to the onChange, so the following should work:
```
onChange={relationships => {
    setFormSettings({ 
        ...formSettings,
        relationships
      });
  }}

```

src/pages/fieldManagement/settings/defaultFieldComponents/RelationshipEditorCore.jsx
```diff
@@ -0,0 +1,239 @@
...
+      )}
+    </div>
+  );
+}
+
+export default function RelationshipEditorCore({
+  schema,
+  value,
+  onChange,
+  siteSettings, // unpacking to prevent passing to div
```
Emily-Ke, line 194: If you are not going to use this prop, is there a reason to pass it in?

Atticus29: > 

The one that live at points older than when I merged in develop I think failed when I tried to rebase them. I'm also worried about re-writing that much of our history at this point.

I'll experiment with this in a backup branch or two tomorrow.

Atticus29:
> I expect that you plan to handle this, but please actually squash the `squash me` commits into something with a meaningful name before this is merged.

> 

The commits that live at points older than when I merged in develop I think failed when I tried to rebase them. I'm also worried about re-writing that much of our history at this point.

I'll experiment with this in a backup branch or two tomorrow.

src/pages/fieldManagement/settings/defaultFieldComponents/RelationshipEditorCore.jsx
```diff
@@ -0,0 +1,239 @@
...
+      </div>
+      <div
+        style={{
+          display: 'flex',
+          flexDirection: 'column',
+          padding: 20,
+        }}
+      >
+        {Object.keys(justRelationshipsFormPart).map(key =>
+          renderSingleRelationshipObj(
```
Emily-Ke, line 230: `renderSingleRelationshipObj` should be a component and should be called as one.
At that point, "render" is implied and "Obj" is misleading, so maybe it should be RelationshipTree or RelationshipTreeEditor.

src/pages/fieldManagement/settings/defaultFieldComponents/RelationshipEditorCore.jsx
```diff
@@ -0,0 +1,239 @@
...
+}
+
+function renderSingleRelationshipObj(
+  allRelationships,
+  targetKey,
+  onChange,
+) {
+  const relationships = get(allRelationships, targetKey, []);
+  return (
+    <div style={{ marginTop: 10 }} key={targetKey}>
```
Emily-Ke, line 139: When `renderSingleRelationshipObj` is called as a React component (mentioned in another comment), this div will no longer need a key.

src/pages/fieldManagement/settings/defaultFieldComponents/RelationshipEditorCore.jsx
```diff
@@ -0,0 +1,239 @@
...
+          ),
+        }}
+      />
+    </div>
+  );
+}
+
+function renderSingleRelationshipObj(
+  allRelationships,
+  targetKey,
```
Emily-Ke, line 134: This is minor, but maybe targetKey should be renamed to something clearer such as category or categoryName.

src/pages/fieldManagement/settings/defaultFieldComponents/RelationshipEditorCore.jsx
```diff
@@ -0,0 +1,239 @@
...
+                    deleteCategory(allRelationships, targetKey),
+                  );
+                }}
+              />
+            </InputAdornment>
+          ),
+        }}
+      />
+      {relationships.map(relationship =>
+        renderSingleRelationship(
```
Emily-Ke, line 179: `renderSingleRelationship` should be a component and should be called as one.
Name ideas include RelationshipRole or RelationshipRoleEditor.

src/pages/fieldManagement/settings/defaultFieldComponents/RelationshipEditorCore.jsx
```diff
@@ -0,0 +1,239 @@
...
+  allRelationships,
+  currentCategory,
+  relationship,
+) {
+  const targetRelationshipCategory = get(
+    allRelationships,
+    currentCategory,
+    [],
+  );
+  const relationshipsWithoutTarget = targetRelationshipCategory.filter(
```
Emily-Ke, line 28: This assumes no duplicate relationships, which is not a safe assumption at this point since deduplication does not happen until later. If I have relationships: ['relationship 1', 'relationship 2', 'relationship 1'] and I delete a row with 'relationship 1' in it, both entries will be deleted.

Emily-Ke (commented):
We should define what we call the different layers of the relationship object returned from houston, so that we can give components and variables meaningful names. What about Relationship > Category (or Type?) > Role?

PR #283, opened 2022-03-21
update individual search API endpoints

Emily-Ke:
With the addition of Elasticsearch to houston in wildmeorg/houston#521, the `/api/v1/search/individuals` endpoint no longer exists. Eventually, we should use the new endpoint, `/api/v1/individuals/search`. However, that will also require changes to the request and response. So, for now, I am changing to `/api/v1/search/proxy/individuals`, which should function exactly the same as `/api/v1/search/individuals`.


brmscheiner (approved):

Atticus29 (approved):
LGTM

PR #285, opened 2022-03-22
DEX-803 make view buttons go to collaboration table directly

Atticus29:
Addresses JH's QA feedback for DEX-803:

> Following up on our Slack conversation. I think all that is left is the make the notifications for a collaboration link to the collaboration table when the “View” button is clicked.

I tried to follow Emily's pattern from her recent PR and use useLocation. I don't think that it fit exactly my use case.
Instead, I followed some online advice(https://dev.to/gedalyakrycer/5-remarkable-react-router-features-anchor-links-query-params-more-2aeg#anchorLinks and https://stackoverflow.com/questions/40280369/use-anchors-with-react-router) and added the 'react-router-hash-link' package.

Slightly added to the ButtonLink component to make the use of HashLinks from within buttons easier going forward.

### DIFF
--- a/src/components/ButtonLink.jsx
+++ b/src/components/ButtonLink.jsx
@@ -1,4 +1,5 @@
 import React from 'react';
+import { HashLink } from 'react-router-hash-link';
 import Button from './Button';
 import Link from './Link';
 
@@ -7,9 +8,21 @@ export default function ButtonLink({
   href,
   external = false,
   newTab = false,
+  isHashLink = false,
   linkProps,
   ...rest
 }) {
+  if (isHashLink) {
+    return (
+      <HashLink
+        style={{ textDecoration: 'unset' }}
+        to={href}
+        {...linkProps}
+      >
+        <Button {...rest}>{children}</Button>
+      </HashLink>
+    );
+  }
   return (
     <Link
       noUnderline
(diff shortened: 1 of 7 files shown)


src/components/UserProfile.jsx
```diff
@@ -174,7 +174,12 @@ export default function UserProfile({
               someoneElse ? 'NO_SIGHTINGS_NON_SELF' : 'NO_SIGHTINGS'
             }
           />
-          {!someoneElse && <CollaborationsCard userId={userId} />}
+          {!someoneElse && (
+            <CollaborationsCard
+              htmlId="collab-card"
```
Atticus29, line 179: Oh yeah. I forgot that I also propagated the necessary element id attribute down to the material element.

src/components/ButtonLink.jsx
```diff
@@ -7,9 +8,21 @@ export default function ButtonLink({
...
+  isHashLink = false,
   linkProps,
   ...rest
 }) {
+  if (isHashLink) {
+    return (
+      <HashLink
+        style={{ textDecoration: 'unset' }}
+        to={href}
+        {...linkProps}
```
Emily-Ke, line 20: It doesn't seem to hurt anything, but HashLink is probably not expecting these props.

Emily-Ke (approved):
One minor comment.

brmscheiner (approved):
very nice!

PR #287, opened 2022-03-24
DEX-720 edit individual metadata default fields

Emily-Ke:
- On the individuals page, reads the metadata values on the metadata card.
- Enables editing of the individual metadata. If the individual does not have a nickname, and none is added, the nickname is not sent as part of the PATCH request.
- Adds validation for required fields to the edit individual metadata form. 
- Removes status as a default field. After discussion, it was decided that users can add it as a custom field if they want to track status. 

I left the names as defaultName/Name and nickname/Nickname(s) for now.

### DIFF
--- a/src/pages/individual/EditIndividualMetadata.jsx
+++ b/src/pages/individual/EditIndividualMetadata.jsx
@@ -149,20 +149,22 @@ export default function EditIndividualMetadata({
             // validation
             const requiredFieldErrors = metadata.reduce(
               (memo, schema) => {
-                if (!schema?.required) return memo;
+                // TODO: Once custom fields can be edited, include them in validation
+                if (
+                  schema &&
+                  (!schema.required || schema.customField)
+                )
+                  return memo;
 
-                const formObject = schema.customField
-                  ? customFieldValues
-                  : defaultFieldValues;
+                if (!defaultFieldValues[schema.name]) {
+                  const fieldName = schema.labelId
+                    ? intl.formatMessage({ id: schema.labelId })
+                    : schema.label;
 
-                if (!formObject[schema.name]) {
-                  const intlFieldName = intl.formatMessage({
-                    id: schema.labelId,
-                  });
                   memo.push(
                     intl.formatMessage(
                       { id: 'IS_REQUIRED' },
-                      { field: intlFieldName },
+                      { field: fieldName },
                     ),
                   );
                 }
--- a/src/pages/individual/EditIndividualMetadata.jsx
+++ b/src/pages/individual/EditIndividualMetadata.jsx
@@ -1,5 +1,5 @@
 import React, { useState } from 'react';
-import { get, set, pick } from 'lodash-es';
+import { get, set } from 'lodash-es';
 
 import DialogContent from '@material-ui/core/DialogContent';
 import DialogActions from '@material-ui/core/DialogActions';
@@ -9,6 +9,13 @@ import CustomAlert from '../../components/Alert';
 import InputRow from '../../components/fields/edit/InputRow';
 import Button from '../../components/Button';
 import StandardDialog from '../../components/StandardDialog';
+
+function deriveNameGuid(metadata, schemaName) {
+  const foundSchema = metadata.find(
+    schema => schema.name === schemaName,
+  );
+  return foundSchema?.nameGuid;
+}
 
 function getInitialFormValues(schema, fieldKey) {
   return schema.reduce((memo, field) => {
@@ -123,12 +130,49 @@ export default function EditIndividualMetadata({
           loading={loading}
           display="primary"
           onClick={async () => {
-            const currentPatchableValues = pick(defaultFieldValues, [
-              'names',
-            ]);
+            const names = [];
+            // Always include the defaultName. If it does not already exist, add it.
+            // Otherwise, replace it.
+            const defaultNameGuid = deriveNameGuid(
+              metadata,
+              'defaultName',
+            );
+            const defaultNameProperty = defaultNameGuid
+              ? {
+                  op: 'replace',
+                  value: defaultFieldValues.defaultName,
+                  guid: defaultNameGuid,
+                }
+              : {
+                  op: 'add',
+                  value: defaultFieldValues.defaultName,
+                  context: 'defaultName',
+                };
+            names.push(defaultNameProperty);
+
+            // If there was a nickname, update it with whatever the value is now.
+            // If there was not a nickname, but there is a value now, add it.
+            // If a nickname did not already exist, don't add it.
+            const nicknameGuid = deriveNameGuid(metadata, 'nickname');
+            const nicknameFieldValue = defaultFieldValues.nickname;
+            if (nicknameGuid) {
+              names.push({
+                op: 'replace',
+                value: nicknameFieldValue,
+                guid: nicknameGuid,
+              });
+            } else if (nicknameFieldValue) {
+              names.push({
+                op: 'add',
+                value: nicknameFieldValue,
+                context: 'nickname',
+              });
+            }
+
+            const properties = { sex: defaultFieldValues.sex, names };
             const successfulUpdate = await updateIndividualProperties(
               individualId,
-              currentPatchableValues,
+              properties,
             );
             if (successfulUpdate) {
               refreshIndividualData();
--- a/src/pages/individual/EditIndividualMetadata.jsx
+++ b/src/pages/individual/EditIndividualMetadata.jsx
@@ -124,7 +124,9 @@ export default function EditIndividualMetadata({
           <CustomAlert severity="error" titleId="SUBMISSION_ERROR">
             {formErrors.length > 0 &&
               formErrors.map(formError => (
-                <Text variant="body2">{formError}</Text>
+                <Text key={formError} variant="body2">
+                  {formError}
+                </Text>
               ))}
             {patchIndividualError && (
               <Text variant="body2">{patchIndividualError}</Text>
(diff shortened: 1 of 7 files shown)


src/pages/individual/EditIndividualMetadata.jsx
```diff
@@ -123,12 +146,78 @@ export default function EditIndividualMetadata({
...
+                if (!schema?.required) return memo;
+
+                const formObject = schema.customField
+                  ? customFieldValues
+                  : defaultFieldValues;
+
+                if (!formObject[schema.name]) {
+                  const intlFieldName = intl.formatMessage({
+                    id: schema.labelId,
+                  });
```
brmscheiner, line 161: custom fields will not have a `labelId`, so we will need to fall back to the `label` property for those

src/pages/individual/EditIndividualMetadata.jsx
```diff
@@ -123,12 +146,78 @@ export default function EditIndividualMetadata({
...
+                  op: 'replace',
+                  value: defaultFieldValues.defaultName,
+                  guid: defaultNameGuid,
+                }
+              : {
+                  op: 'add',
+                  value: defaultFieldValues.defaultName,
+                  context: 'defaultName',
+                };
+            names.push(defaultNameProperty);
```
brmscheiner, line 196: might be a little cleaner to declare `names` at this point instead
```
const names = [defaultNameProperty] 
```

src/pages/individual/EditIndividualMetadata.jsx
```diff
@@ -123,12 +146,78 @@ export default function EditIndividualMetadata({
...
+            const nicknameFieldValue = defaultFieldValues.nickname;
+            if (nicknameGuid) {
+              names.push({
+                op: 'replace',
+                value: nicknameFieldValue,
+                guid: nicknameGuid,
+              });
+            } else if (nicknameFieldValue) {
+              names.push({
+                op: 'add',
```
brmscheiner, line 211: i'm pretty sure you can use `op: 'replace'` in either situation and spare the complexity. same with defaultName above

src/pages/individual/Individual.jsx
```diff
@@ -46,30 +50,56 @@ export default function Individual() {
...
+      deriveIndividualName(
+        individualData,
+        'defaultName',
+        'Unnamed individual',
+      ),
+      deriveIndividualName(
+        individualData,
+        'nickname',
+        'Unnamed individual',
+      ),
```
brmscheiner, line 99: if there is no nickname, let's just leave off the nickname text rather than call it "unnamed individual"

src/pages/individual/Individual.jsx
```diff
@@ -192,7 +222,11 @@ export default function Individual() {
             )}
             assets={fakeAssets}
           />
-          <MetadataCard editable metadata={metadata} />
+          <MetadataCard
+            editable
```
brmscheiner, line 226: @colinwkingen is there a way to tell whether the currently logged in user is able to edit an individual's metadata?

brmscheiner (approved):

src/pages/individual/EditIndividualMetadata.jsx
```diff
@@ -9,6 +10,14 @@ import CustomAlert from '../../components/Alert';
 import InputRow from '../../components/fields/edit/InputRow';
 import Button from '../../components/Button';
 import StandardDialog from '../../components/StandardDialog';
+import Text from '../../components/Text';
+
+function deriveNameGuid(metadata, schemaName) {
+  const foundSchema = metadata.find(
+    schema => schema.name === schemaName,
+  );
+  return foundSchema?.nameGuid;
```
Atticus29, line 19: What if there's more than one match in the .find?

src/pages/individual/EditIndividualMetadata.jsx
```diff
@@ -104,17 +120,24 @@ export default function EditIndividualMetadata({
...
 
-        {error && (
+        {showErrorAlert && (
           <CustomAlert severity="error" titleId="SUBMISSION_ERROR">
-            {error}
+            {formErrors.length > 0 &&
+              formErrors.map(formError => (
+                <Text variant="body2">{formError}</Text>
+              ))}
+            {patchIndividualError && (
```
Atticus29, line 129: I find myself wondering whether it's possible to just add the patchIndividual error to the formErrors array.

src/pages/individual/EditIndividualMetadata.jsx
```diff
@@ -123,12 +146,78 @@ export default function EditIndividualMetadata({
...
           onClick={async () => {
-            const currentPatchableValues = pick(defaultFieldValues, [
-              'names',
-            ]);
+            // validation
+            const requiredFieldErrors = metadata.reduce(
+              (memo, schema) => {
+                if (!schema?.required) return memo;
+
+                const formObject = schema.customField
```
Atticus29, line 154: schema?.customField

src/pages/individual/EditIndividualMetadata.jsx
Atticus29: That seems like a lot of extra work for very little payoff, though.

src/pages/individual/Individual.jsx
```diff
@@ -46,30 +50,56 @@ export default function Individual() {
...
-  function refreshIndividualData() {
-    const queryKey = getIndividualQueryKey(id);
-    queryClient.invalidateQueries(queryKey);
-  }
+  const refreshIndividualData = useCallback(
+    () => {
+      const queryKey = getIndividualQueryKey(id);
+      queryClient.invalidateQueries(queryKey);
+    },
+    [queryClient],
```
Atticus29, line 58: When would the queryClient change?

src/utils/nameUtils.js
```diff
@@ -1,9 +1,21 @@
+function getNameObject(individualData, context) {
+  const names = individualData?.names || [];
+  return names.find(n => n?.context === context) || {};
```
Atticus29, line 3: Return an array or an object? Intentional?

Atticus29 (commented):
Small things!

src/pages/individual/EditIndividualMetadata.jsx
brmscheiner: I think find takes the first matching item

Emily-Ke: Yes, it returns the first value that satisfies the test, or, if no value does, it returns `undefined`.

src/pages/individual/Individual.jsx
Emily-Ke: I don't think it ever would. We set the queryClient instance once in App.
However, according to the [React documentation](https://reactjs.org/docs/hooks-reference.html#usecallback):
> The array of dependencies is not passed as arguments to the callback. Conceptually, though, that’s what they represent: every value referenced inside the callback should also appear in the dependencies array. In the future, a sufficiently advanced compiler could create this array automatically.
>
> We recommend using the exhaustive-deps rule as part of our eslint-plugin-react-hooks package. It warns when dependencies are specified incorrectly and suggests a fix.

When I set up the exhaustive-deps rule, and removed queryClient from the array, it warned:
```
/codex-frontend/src/pages/individual/individual.jsx
  58:5  warning  React Hook useCallback has missing dependencies: 'id' and 'queryClient'. Either include them or remove the dependency array  react-hooks/exhaustive-deps
 ``` 
**TODO**: Add `id` to the dependency array too.

src/utils/nameUtils.js
Emily-Ke: `names` is an array of name objects. `names.find` should return either the name object with the matching context or `undefined`. So, this function should return either a name object or an empty object.

Atticus29: Sorry. I thought that find was like map and returned an array. Thanks!

src/pages/individual/EditIndividualMetadata.jsx
Emily-Ke: Addressed in https://github.com/WildMeOrg/codex-frontend/pull/287/commits/5ffb629e36b90928abb8b618ba8b77e53bb17914

I added `schema.label` as the fallback if `schema.labelId` is missing, but I also removed validation for custom field values because editing custom field values will be handled in a separate PR.

src/pages/individual/EditIndividualMetadata.jsx
Emily-Ke: Fixed in https://github.com/WildMeOrg/codex-frontend/pull/287/commits/3c4e9c3a4c2dd44627eec3fe3a36946dd7e8c0ed

src/pages/individual/EditIndividualMetadata.jsx
Emily-Ke: I don't think you can. `op: 'replace'` requires a `guid`, which won't exist if the name does not already exist.
https://github.com/WildMeOrg/houston/blob/main/app/modules/individuals/parameters.py#L196-L207

src/pages/individual/Individual.jsx
Emily-Ke: There is already a separate ticket for this (DEX-844), so I'm leaving this as is, since it preserves the current behavior.

src/utils/nameUtils.js
Emily-Ke: You might be thinking of filter.

src/pages/individual/EditIndividualMetadata.jsx
brmscheiner: I think there is a distinction between a formError and a requestError. From my perspective it's ok to keep them separate!

src/pages/individual/EditIndividualMetadata.jsx
brmscheiner: you can supply the guid though. but either way is fine

PR #290, opened 2022-03-25
Refactor relationship types

brmscheiner:
I updated the schema as discussed in chat. As a result, you might not be able to load the UI until you clear out your existing relationship_type in site settings. Using postman to set it to `{}` should do the trick

UI is more or less the same but I did add some validation and error handling to replace the alert

image


src/pages/fieldManagement/settings/defaultFieldComponents/RelationshipEditor.jsx
```diff
@@ -0,0 +1,114 @@
...
+}
+
+function validateRelationships(relationships) {
+  let errors = [];
+  const types = Object.values(relationships);
+  const typeLabels = types.map(t => t?.label);
+  const emptyTypeLabels = typeLabels.filter(label => label === '');
+  if (emptyTypeLabels.length > 0)
+    errors.push('One or more relationship types are missing labels.');
+  const uniqueTypeLabels = uniq(typeLabels);
```
Atticus29, line 26: uniq! Cool!

src/pages/fieldManagement/settings/defaultFieldComponents/relationshipComponents/Type.jsx
```diff
@@ -0,0 +1,86 @@
...
+import InputAdornment from '@material-ui/core/InputAdornment';
+import IconButton from '@material-ui/core/IconButton';
+import NewChildIcon from '@material-ui/icons/AddCircle';
+
+import TextInput from '../../../../../components/inputs/TextInput';
+import DeleteButton from '../../../../../components/DeleteButton';
+import Role from './Role';
+
+function deleteType(relationships, typeGuid) {
+  return omit(relationships, [typeGuid]);
```
Atticus29, line 14: Doesn't this suffer from the same within-object editing issue that we were seeing on my PR?

src/pages/fieldManagement/settings/defaultFieldComponents/relationshipComponents/Type.jsx
```diff
@@ -0,0 +1,86 @@
...
+}
+
+function updateType(relationships, typeGuid, newTypeLabel) {
+  const type = get(relationships, typeGuid, {});
+  return {
+    ...relationships,
+    [typeGuid]: {
```
Atticus29, line 21: So, if relationships already has a property with key typeGuid, it will simply be overwritten by the subsequent
[typeGuid]: {...}
?
If so, that's very cool!

src/pages/fieldManagement/settings/defaultFieldComponents/relationshipComponents/Type.jsx
```diff
@@ -0,0 +1,86 @@
...
+      ...type,
+      label: newTypeLabel,
+    },
+  };
+}
+
+function addRole(relationships, typeGuid) {
```
Atticus29, line 28: Validation rather than post-hoc de-duping was definitely the cleaner choice here. Nice.

src/pages/fieldManagement/settings/defaultFieldComponents/relationshipComponents/Type.jsx
```diff
@@ -0,0 +1,86 @@
...
+  const type = get(relationships, typeGuid, {});
+  const currentRoles = type?.roles || [];
+  return {
+    ...relationships,
+    [typeGuid]: {
+      ...type,
```
Atticus29, line 34: again, roles overwrites the one that will exist in ...type?

src/pages/fieldManagement/settings/defaultFieldComponents/relationshipComponents/Type.jsx
Atticus29, line 28: I totally see why it is structured this way, but it's slightly weird that addRole lives in Types.jsx, but all of the other role-modifying methods live in Roles.jsx.

Atticus29 (approved):
This is awesome.
My big take home is that you can edit something that gets spread in a .map call by subsequently including that same key of that this in the same map call. That's suuper helpful.
Well done!

src/pages/fieldManagement/settings/defaultFieldComponents/relationshipComponents/Type.jsx
brmscheiner: it says in the lodash docs that omit returns a brand new object: "this method creates an object composed of the own and inherited enumerable property paths of object that are not omitted."

src/pages/fieldManagement/settings/defaultFieldComponents/relationshipComponents/Type.jsx
brmscheiner: that's it! with object spreading the items at the end of the statement overwrite the items at the beginning.

src/pages/fieldManagement/settings/defaultFieldComponents/relationshipComponents/Type.jsx
brmscheiner: yeah, I just moved them where they were used. it's on the type component where the add role button actually lives :)

src/pages/fieldManagement/settings/defaultFieldComponents/RelationshipEditor.jsx
```diff
@@ -0,0 +1,114 @@
...
+export default function RelationshipEditor({
+  onClose,
+  onSubmit,
+  formSettings,
+  setFormSettings,
+}) {
+  const [formErrors, setFormErrors] = useState(null);
+
+  function setRelationships(relationships) {
+    setFormSettings({ relationships });
```
Emily-Ke, line 57: This should be `setFormSettings({ ...formSettings, relationships });`.

As @Atticus29 said earlier, it doesn't seem like this form state needs to be stored farther up the component tree than the `RelationshipEditor`. Maybe that is something to consider changing in the future.
>I'm guessing that this has something to do with all of them belonging to the same form schema. IMHO they don't need to be.

_Originally posted by @Atticus29 in https://github.com/WildMeOrg/codex-frontend/pull/282#discussion_r833771925_

Emily-Ke (changes requested):
Because you cleaned up how relationship changes are submitted, there is some duplication in DefaultFieldTable's editField.Editor's onSubmit. It might be worth abstracting.

src/pages/fieldManagement/settings/defaultFieldComponents/RelationshipEditor.jsx
```diff
@@ -0,0 +1,114 @@
...
+  if (emptyTypeLabels.length > 0)
+    errors.push('One or more relationship types are missing labels.');
+  const uniqueTypeLabels = uniq(typeLabels);
+  if (uniqueTypeLabels.length !== typeLabels.length)
+    errors.push(
+      'Two or more relationship types have the same label. Make sure each label is different.',
+    );
+  types.forEach(type => {
+    const typeLabel = type?.label;
+    const roleLabels = get(type, 'roles', []).map(r => r?.label);
```
Emily-Ke, line 33: Perhaps add a validation that the type has at least one role label.

src/pages/fieldManagement/settings/defaultFieldComponents/RelationshipEditor.jsx
```diff
@@ -31,7 +31,11 @@ function validateRelationships(relationships) {
   types.forEach(type => {
     const typeLabel = type?.label;
     const roleLabels = get(type, 'roles', []).map(r => r?.label);
-    const emptyRoleLabels = roleLabels.filter(label => label === '');
+    if (roleLabels.length === 0)
+      errors.push(
+        `Each relationship type must have at least one role.`,
+      );
```
Emily-Ke, line 37: If two different types each have no roles, this error message is printed twice. Because the error messages are also used as keys, the duplicated messages also cause a key warning.

Atticus29: I noticed this. I think the errors (which mention the types explicitly) do a good job clarifying where the end user should be looking for the error(s). As for key warning, maybe make the key the type?

Emily-Ke: The "each relationship type must have at least one role." error prints twice. I favor either only printing that particular message once or adding the type information like you did for the other errors: "Type ${typeLabel} must have at least one role."
duplicate-error-messages

Emily-Ke (approved):

PR #291, opened 2022-03-29
DEX-846 comment out subscribe button from Individual page. Was alread…

Atticus29:
…y commented out in in sighting and asset group


brmscheiner (approved):
Really really really good work on this ticket

PR #294, opened 2022-03-31
DEX-862 update individual name contexts

Emily-Ke:
Updates the name contexts and labels from defaultName (Name) to FirstName (First name) and from nickname (Nickname(s)) to AdoptionName (Adoption name).


src/models/individual/useIndividualFieldSchemas.js
```diff
@@ -38,23 +38,23 @@ export default function useIndividualFieldSchemas() {
...
-          labelId: 'NAME',
+          name: 'firstName',
+          labelId: 'FIRST_NAME',
           category: defaultIndividualCategories.general.name,
           requiredForIndividualCreation: true,
           required: true,
           defaultValue: '',
           getValue: (_, individualData) =>
-            deriveIndividualName(individualData, 'defaultName', ''),
+            deriveIndividualName(individualData, 'FirstName', ''),
```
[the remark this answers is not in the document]
Atticus29: Maybe this is back end thing or a schema convention I'm not aware of?

src/pages/individual/EditIndividualMetadata.jsx
```diff
@@ -179,42 +179,46 @@ export default function EditIndividualMetadata({
...
-            if (nicknameGuid) {
+            // If there was an adoptionName, update it with whatever the value is now.
+            // If there was not an adoptionName, but there is a value now, add it.
+            // If an adoptionName did not already exist, don't add it.
+            const adoptionNameGuid = deriveNameGuid(
+              metadata,
+              'AdoptionName',
+            );
+            const adoptionNameFieldValue =
+              defaultFieldValues.adoptionName;
```
Atticus29, line 210: I didn't quite follow how this is a guaranteed non-null?

src/utils/notificationUtils.js
```diff
@@ -1,26 +1,29 @@
...
   const user1Name =
     notification?.message_values?.user1_name || 'Unnamed User';
   const user2Name =
     notification?.message_values?.user2_name || 'Unnamed User';
-  const individual1Names = notification?.names || []; // TODO flesh out more once this is included in notifications DEX-739
-  const individual1NicknameObject = individual1Names.find(
-    n => n.context === 'nickname',
+
+  // TODO flesh out more once this is included in notifications DEX-739
+  const individual1AdoptionName = deriveIndividualName(
```
Atticus29, line 11: thank you for making this use deriveIndividualName

Emily-Ke: The blocker for DEX-739 appears to have been removed, so at some point we should get back to this and not set both names to the same value. 😆

brmscheiner (approved):

src/pages/individual/EditIndividualMetadata.jsx
Emily-Ke: The way I see it, defaultFieldValues is initially set on declaration to the return value from getInitialFormValues, which will be an object. defaultFieldValues is only ever reassigned in the onChange for the edit component. The value that it is set to is at least an empty object. I don't see a way that defaultFieldValues could ever be null, but I may have missed something.

PR #295, opened 2022-03-31
Dex 844 fix individual header

Atticus29:
Only displays, "also known as" when adoptionName is non-null.

Allows for user to both display and edit the featuredAsset of the individual page without stepping on the toes of editing the featuredAsset from the sighting page.

I decided to keep the shared component (otherwise, what is the point of modular components??) and transform the data going into the FeaturedPhoto component on the Individual.jsx end.

Confirmed that this works when more than one encounter is attached to the individual.


src/pages/individual/Individual.jsx
```diff
@@ -58,6 +58,36 @@ export default function Individual() {
...
+    () => {
+      const allAssets = reduce(
+        individualData?.encounters,
+        (memo, encounter) => {
+          const newAssets = map(
+            get(encounter, 'annotations', []),
+            annotation => ({
+              src: annotation?.asset_src,
+              guid: annotation?.asset_guid,
+              filename: annotation?.created
```
Atticus29, line 71: Happy to no longer call this filename in favor of something like altText if that is desired.

src/pages/sighting/featuredPhoto/FeaturedPhotoSelector.jsx
```diff
@@ -69,28 +79,37 @@ export default function FeaturedPhotoSelector({
           flexDirection: 'column',
         }}
       >
-        {error && (
+        {(sightingError || individualError) && (
```
Atticus29, line 82: Figured it was small enough.

locale/en.json
```diff
@@ -997,7 +997,9 @@
   "BOTH_VULNERABLE_MESSAGE": "Deleting this encounter would result in assigned individuals and/or parent sightings being deleted. Are you sure you want to continue?",
   "DETACH_ENCOUNTER_FROM_INDIVIDUAL": "Detach this sighting from this individual",
   "DETACH_ENCOUNTER_FROM_INDIVIDUAL_DISABLED": "Cannot detach the final sighting from this individual",
+  "DUPLICATE_RELATIONSHIPS_FOUND_AND_REMOVED": "Duplicate relationships in one or more categories identified and removed",
+  "UNNAMED_INDIVIDUAL": "Unnamed individual",
+  "FEATURED_ASSET": "Featured asset",
```
brmscheiner, line 1002: let's use "Top photograph" for user-facing stuff

src/pages/individual/Individual.jsx
brmscheiner: Definitely, thanks.

src/pages/individual/Individual.jsx
```diff
@@ -58,6 +58,36 @@ export default function Individual() {
...
+                : 'Annotation with unknown creation date',
+            }),
+          );
+          return [...memo, ...newAssets];
+        },
+        [],
+      );
+      const assets = uniqBy(allAssets, asset => asset.src);
+      return {
+        assets: assets,
```
brmscheiner, line 82: You can just do `assets,` and it will work the same way

src/pages/individual/Individual.jsx
```diff
@@ -122,6 +152,11 @@ export default function Individual() {
   const [merging, setMerging] = useState(false);
   const [deletingIndividual, setDeletingIndividual] = useState(false);
   const [deleteEncounterId, setDeleteEncounterId] = useState(null);
+  const adoptionNameIsLegitimate =
```
brmscheiner, line 155: Instead of doing this, why not just omit the fallback when deriving adoptionName. then below you can just do `{adoptionName && ...}`

src/pages/sighting/featuredPhoto/FeaturedPhoto.jsx
```diff
@@ -1,5 +1,5 @@
 import React, { useState } from 'react';
```
brmscheiner, line 1: Now that we're using this component in multiple areas of the app we should move it to `src/components`

src/pages/sighting/featuredPhoto/FeaturedPhotoSelector.jsx
```diff
@@ -9,21 +8,32 @@ import CustomAlert from '../../../components/Alert';
...
+  const {
+    updateProperties,
+    loading: sightingLoading,
+    error: sightingError,
+  } = usePatchSighting();
+  const {
+    updateIndividualProperties,
+    loading: individualLoading,
+    error: individualError,
+  } = usePatchIndividual();
```
brmscheiner, line 36: Another option would be to pass the mutation in as a prop. This way we'll have to continue adding `usePatchXYZ...` to this file whenever the component gets used for another entity type

brmscheiner (approved):
Approved, some cleanup/reorganization related comments that you can do if you want

src/pages/individual/Individual.jsx
Atticus29: FeaturedPhotos or whatever is looking for get(data, 'assets', [])... how would that work the same way?

src/pages/individual/Individual.jsx
Atticus29: Great point.

src/pages/sighting/featuredPhoto/FeaturedPhotoSelector.jsx
Atticus29: Good idea!

locale/en.json
Atticus29: Done

src/pages/individual/Individual.jsx
Atticus29: What does "just do assets" in this case mean?

src/pages/individual/Individual.jsx
Atticus29: Done

src/pages/individual/Individual.jsx
Emily-Ke: `{ assets: assets }` can be written as `{ assets }`. It is just a shorthand.

Atticus29: I could have sworn that I tried it this way. I will confirm. Thanks.

src/pages/sighting/featuredPhoto/FeaturedPhoto.jsx
Atticus29: Done.

src/pages/sighting/featuredPhoto/FeaturedPhoto.jsx
```diff
@@ -39,23 +40,23 @@ export default function FeaturedPhoto({
     >
       <FeaturedPhotoSelector
         currentFeaturedPhotoId={featuredPhotoGuid}
-        sightingId={get(data, 'id')}
+        sightingId={
+          !data?.isFromIndividual ? get(data, 'guid') : null
+        }
+        individualId={
+          data?.isFromIndividual ? get(data, 'guid') : null
+        }
```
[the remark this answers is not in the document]
Atticus29: Done

src/pages/sighting/featuredPhoto/FeaturedPhotoSelector.jsx
Atticus29: Respectfully, gonna punt this until we hit the scenario where we're gonna have to use this in another entity type and cross that bridge when we get there.

src/pages/individual/Individual.jsx
Atticus29: Done

src/pages/individual/Individual.jsx
Atticus29: Done

src/components/FeaturedPhotoSelector.jsx
```diff
@@ -1,29 +1,39 @@
...
+import StandardDialog from './StandardDialog';
+import usePatchIndividual from '../models/individual/usePatchIndividual';
 
 export default function FeaturedPhotoSelector({
   open,
   onClose,
   assets,
-  sightingId,
+  sightingId = null,
+  individualId = null,
```
Emily-Ke, line 18: `sightingId` and `individualId` will be undefined if they are not passed as properties on the props object. Why do you want to set them specifically to `null`?

src/components/FeaturedPhotoSelector.jsx
```diff
@@ -69,28 +79,37 @@ export default function FeaturedPhotoSelector({
...
         }}
       >
-        {error && (
+        {(sightingError || individualError) && (
           <CustomAlert
             style={{ marginBottom: 20 }}
             severity="error"
             titleId="ERROR_UPDATING_PROFILE"
-            description={error}
+            description={sightingId ? sightingError : individualError}
```
Emily-Ke, line 87:
```
const error = sightingError || individualError;
if (error) {
  // ...
  description={error}
}
```
This way the error that causes the custom alert to render is the one that is displayed.

Emily-Ke (approved):
A couple of minor comments.

src/components/FeaturedPhotoSelector.jsx
Atticus29: Didn't realize that I could optionally neglect props that didn't have default values. Cool!

src/components/FeaturedPhotoSelector.jsx
Atticus29: Done

src/components/FeaturedPhotoSelector.jsx
Atticus29: Done

PR #296, opened 2022-03-31
DEX-862 fix individual name update

Emily-Ke:
Updating an existing name was resulting in an error. Because the incorrect schema names were being passed to `deriveNameGuid`, no guids were found, and the operation in the PATCH request to houston was being set as "add" instead of "replace". This bug was introduced in #294. 

As a part of the fix, I looked into simplifying the request to only use replace and send a guid if it didn't already exist. However, houston throws an exception if the guid in the request does not match an existing name.
https://github.com/WildMeOrg/houston/blob/1ce45296747b4d3becd1f794e8e25053c80f0f33/app/modules/individuals/parameters.py#L210-L216


Atticus29 (approved):
LGTM

PR #299, opened 2022-04-01
DEX-869 disable continuing without assets if user does not have a res…

Atticus29:
…earcher role

### DIFF
--- a/src/pages/reportSighting/ReportSighting.jsx
+++ b/src/pages/reportSighting/ReportSighting.jsx
@@ -14,9 +14,15 @@ import Link from '../../components/Link';
 import Button from '../../components/Button';
 import UppyDashboard from '../../components/UppyDashboard';
 import ReportForm from './ReportForm';
+import useGetMe from '../../models/users/useGetMe';
 
 export default function ReportSighting({ authenticated }) {
   const intl = useIntl();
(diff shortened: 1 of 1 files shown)


src/pages/reportSighting/ReportSighting.jsx
```diff
@@ -14,16 +14,28 @@ import Link from '../../components/Link';
...
 import UppyDashboard from '../../components/UppyDashboard';
 import ReportForm from './ReportForm';
+import useGetMe from '../../models/users/useGetMe';
 
 export default function ReportSighting({ authenticated }) {
   const intl = useIntl();
+  const {
+    data: currentUserData,
+    loading: userDataLoading,
+    isFetching: userDataFetching,
```
Emily-Ke, line 24: I'm not sure that it is necessary to watch and react to these states. This component is a child of FrontDesk and won't even be rendered if there is no data.

Emily-Ke (approved):
Looks good to me.

src/pages/reportSighting/ReportSighting.jsx
Atticus29: Great point. Removed.

brmscheiner (approved):

PR #300, opened 2022-04-01
DEX-745 encounter metadata - edit custom fields

Emily-Ke:
- Adds the customFields object to the PATCH request when submitting edits to an encounter's metadata.
- Adds validation for required fields, both custom and default.


Atticus29 (approved):
Makes sense to me! Nice!

brmscheiner (approved):

PR #301, opened 2022-04-04
Dex 822 fix scroll

Atticus29:


Emily-Ke (commented):
This fixes part of the problem, but the maxHeight prop may need to be supplied a different value for some of the Card instances.

brmscheiner:
^agreed with @Emily-Ke. Also i found the original ticket that caused me to do overflow: scroll

https://wildme.atlassian.net/browse/DEX-797

Can you verify that this doesn't cause a regression for that issue? Testing multiple browsers could be good too, I never saw any of this stuff in chrome/brave

Atticus29:
>

Atticus29:
> ^agreed with @Emily-Ke. Also i found the original ticket that caused me to do overflow: scroll
> 
> https://wildme.atlassian.net/browse/DEX-797
> 
> Can you verify that this doesn't cause a regression for that issue? Testing multiple browsers could be good too, I never saw any of this stuff in chrome/brave

Confirmed. Note, however, that both in chrome and firefox, I have to two-finger scroll to even see the scrollbar there.
This felt very organic to me, but maybe not for all users? I don't know how to make the appearance of the horizontal scrollbar explicit? If that's even desired?

brmscheiner:
Maybe `overflow-x`? Don't think it needs to be explicit as long as it works though

Atticus29:
> 

Should I maybe do like a mouse hover x-overflow: scroll?

brmscheiner (approved):

PR #302, opened 2022-04-05
Dex 863 z indexing of alerts

Atticus29:
The problem was that for annotation creation, the dialog is full screen, so the alert was popping up in a part of the dialog that it shouldn't have been.

### DIFF
--- a/src/components/AnnotationCreator.jsx
+++ b/src/components/AnnotationCreator.jsx
@@ -241,6 +241,8 @@ export default function AnnotationCreator({
               viewpoint,
               theta,
             );
+            console.log('deleteMe newAnnotationId is: ');
+            console.log(newAnnotationId);
             if (newAnnotationId) {
               refreshSightingData();
               onClose();
--- a/src/components/AnnotationCreator.jsx
+++ b/src/components/AnnotationCreator.jsx
@@ -126,15 +126,6 @@ export default function AnnotationCreator({
           />
         </div>
       </DialogContent>
-      {/* {error && (
-        <CustomAlert
-          titleId="SERVER_ERROR"
-          style={{ marginTop: 16, marginBottom: 8 }}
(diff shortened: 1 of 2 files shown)


Emily-Ke (approved):
It's not quite consistent with our other dialogs.

Atticus29:
> 

It's the only <StandardDialog> that has a fullscreen prop that also has an alert, so I kinda just decided what looked best 🤷‍♂️ .

src/components/AnnotationCreator.jsx
```diff
@@ -249,6 +230,15 @@ export default function AnnotationCreator({
...
       </DialogActions>
+      {error && (
+        <CustomAlert
+          titleId="SERVER_ERROR"
+          style={{ marginTop: 16, marginBottom: 8 }}
+          severity="error"
+        >
+          {error}
+        </CustomAlert>
+      )}
```
brmscheiner, line 241: does it look weird for the error to be below the buttons? everywhere else we do the errors above them

Atticus29: Ok... moved it above. It looks better, I agree:
Screen Shot 2022-04-07 at 5 29 34 PM

src/components/AnnotationCreator.jsx
```diff
@@ -207,6 +188,23 @@ export default function AnnotationCreator({
...
         </FormControl>
       </div>
+      {error && (
+        <CustomAlert
+          titleId="SERVER_ERROR"
+          style={{
+            marginTop: 16,
+            marginBottom: 8,
+            marginLeft: 'auto',
+            marginRight: 'auto',
```
brmscheiner, line 198: you can abbreviate this `margin: '16px auto 8px auto'`

brmscheiner (approved):

PR #304, opened 2022-04-07
Fix annotation card scroll

brmscheiner:
Addresses DEX-883

### DIFF
--- a/src/pages/sighting/encounters/AnnotationsCard.jsx
+++ b/src/pages/sighting/encounters/AnnotationsCard.jsx
@@ -61,56 +61,54 @@ export default function AnnotationsCard({
         sightingData={sightingData}
       />
       {annotationReferences.length > 0 ? (
-        <div style={{ margin: '8px 0' }}>
-          <Grid container spacing={2}>
-            {annotationReferences.map(annotationReference => {
-              const matchingAsset = assets.find(
-                asset =>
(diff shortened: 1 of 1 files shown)


src/pages/sighting/encounters/AnnotationsCard.jsx
```diff
@@ -61,54 +61,56 @@ export default function AnnotationsCard({
...
-                    })
-                  }
-                  style={{
-                    position: 'absolute',
-                    top: 12,
-                    right: 12,
-                    color: theme.palette.common.white,
-                  }}
-                  className={classes.photoIcon}
+        <div style={{ margin: '8px 0' }}>
```
brmscheiner, line 64: surrounding the Grid in a div with 8px margin on top and bottom cancels out the -8px from material-ui

Atticus29 (approved):
Good to know. Thanks!

src/pages/sighting/encounters/AnnotationsCard.jsx
Emily-Ke: I think the style could have been added directly to the Grid and saved you one div.

Emily-Ke (approved):

src/pages/sighting/encounters/AnnotationsCard.jsx
brmscheiner: that works! changed

PR #306, opened 2022-04-08
DEX-868 add merge pending screen

Emily-Ke:
Adds a page to redirect users to when they do not have permission to automatically merge individuals.

## New route
`/pending-merges/:guid`

### Loading
merge-pending-loading

### Success (with and without deadline)
merge-pending-success
merge-pending-date-missing

### Error
merge-pending-error


src/pages/merge/IndividualsMergePending.jsx
```diff
@@ -0,0 +1,136 @@
+import React from 'react';
```
brmscheiner, line 1: maybe a few too many components in one file here, otherwise lgtm

brmscheiner (approved):

src/pages/merge/IndividualsMergePending.jsx
```diff
@@ -0,0 +1,136 @@
...
+  const theme = useTheme();
+  const themeColor = theme.palette.primary.main;
+
+  return isError ? (
+    <BaoSpills
+      themeColor={themeColor}
+      style={{ width: baoSpillsWidth, margin: '72px 8px 20px 0' }}
+    />
+  ) : (
+    <BaoConfused
```
Atticus29, line 37: I don't love BaoConfused for an expected user experience. I don't think that this comment is really in the purview of this ticket, but I almost think I'd like BaoDetective better here? I dunno. @brmscheiner would it be crazy to commission a BaoWaiting?

src/pages/merge/IndividualsMergePending.jsx
Atticus29: Agreed. This file feels ever-so-slightly long and might could use some chopping.

Atticus29 (commented):
Same as Ben plus a gripe about the confused Bao that may or may not be an issue for the ticket at hand, but I think worth a discussion at least.

src/pages/merge/IndividualsMergePending.jsx
brmscheiner: dont think we have time to worry about new ones right now. feel free to try baoDetective though and switch it out if it looks better

Emily-Ke:
Switched to BaoLetter in https://github.com/WildMeOrg/codex-frontend/pull/306/commits/44155b11939cd92cb48c0f43297506a2b4b4cec5

merge-pending-letter

src/pages/merge/IndividualsMergePending.jsx
Emily-Ke: Reduced components in https://github.com/WildMeOrg/codex-frontend/pull/306/commits/83eeaef18db77413cc4e8b62bb7a5d5ffac60fa7

Atticus29 (approved):
Looks good!!!

brmscheiner (approved):

PR #307, opened 2022-04-09
DEX-721 individual metadata custom fields

Emily-Ke:
- Enables custom field editing for individuals, including validation of required fields.
- Resets the EditIndividualMetadata form values when the cancel or close buttons are clicked.
  - Updates the close button callback so that the behavior is the same whether the cancel button or the close button is clicked.
- Removes a duplicate entry from `locale/en.json`

### DIFF
--- a/src/pages/individual/EditIndividualMetadata.jsx
+++ b/src/pages/individual/EditIndividualMetadata.jsx
@@ -72,12 +72,24 @@ export default function EditIndividualMetadata({
   const showErrorAlert =
     patchIndividualError || formErrors.length > 0;
 
+  function handleClose() {
+    setDefaultFieldValues(
+      getInitialFormValues(defaultFieldMetadata, 'name'),
+    );
+    setCustomFieldValues(
+      getInitialFormValues(customFieldMetadata, 'id'),
+    );
+    setPatchIndividualError(null);
+    setFormErrors([]);
+    onClose();
+  }
+
   return (
     <StandardDialog
       PaperProps={{ style: { width: 800 } }}
       maxWidth="lg"
       open={open}
-      onClose={onClose}
+      onClose={handleClose}
       titleId="EDIT_INDIVIDUAL_METADATA"
     >
       <DialogContent style={{ minWidth: 200 }}>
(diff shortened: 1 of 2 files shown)


Atticus29 (approved):
Just one question. Otherwise, LGTM!

src/pages/individual/EditIndividualMetadata.jsx
```diff
@@ -135,38 +154,29 @@ export default function EditIndividualMetadata({
...
+                if (!schema.required) return memo;
+
+                const isFieldEmpty = schema.customField
+                  ? !customFieldValues[schema.id]
+                  : !defaultFieldValues[schema.name];
+
+                if (isFieldEmpty) {
                   const fieldName = schema.labelId
                     ? intl.formatMessage({ id: schema.labelId })
                     : schema.label;
```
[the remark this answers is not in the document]
Emily-Ke: Good catch on schema being an issue. I forgot an optional chaining operator on line 165
```
if (!schema.required) return memo;
```

If both labelId and label are null, I don't think anything will fail, but the message will be " is a required field."

src/pages/individual/EditIndividualMetadata.jsx
```diff
@@ -52,32 +60,43 @@ export default function EditIndividualMetadata({
...
+        ),
+      );
+      setCustomFieldValues(
+        getInitialFormValues(getCustomFieldMetadata(metadata), 'id'),
+      );
+      setPatchIndividualError(null);
+      setFormErrors([]);
+      onClose();
+    },
+    [metadata, setPatchIndividualError, onClose],
```
brmscheiner, line 91: not sure `setPatchIndividualError` needs to be included in the dependency array. since this is coming from useState i'm not sure the function will ever change in a meaningful way. was the linter making you do this though? 

i was reading about functions in dependency arrays here: 
https://reacttraining.com/blog/when-to-use-functions-in-hooks-dependency-array/

brmscheiner (approved):

src/pages/individual/EditIndividualMetadata.jsx
Emily-Ke: > When do I use functions in a Hooks Dependency Array?
> The shore answer is...*whenever the linter tells you to*.

```
src/pages/individual/EditIndividualMetadata.jsx
  91:5  warning  React Hook useCallback has a missing dependency: 'setPatchIndividualError'. Either include it or remove the dependency array  react-hooks/exhaustive-deps
```

I do agree that we know that setPatchIndividualError is just the setter function from a useState call so it can be safely omitted

PR #310, opened 2022-04-15
Matching set query builder...

brmscheiner:
Sighting level ID kickoff happens on AGS commit... 
image

Here is what the UI looks like 
image

Annotation ID kickoff is available for sightings only (not AGS) and can be found in the annotations tab
image

Here is what the UI looks like 
image

As you can see the form from the "custom" mode for sighting level ID jobs and annotation ID jobs are identical. Both construct annotation search queries to specify the matching set.

Notes about the current state:
1. Immediately after reporting an AGS, the commit button will briefly appear and then disappear. If you managed to get the dialog open in that amount of time, it will disappear on you as well. This problem will be fixed when the updated status properties are available.
2. There is no way to look at the running job, and no way to view match results. That is next...
3. Eventually would be great to populate the algorithms field with recommended algorithms but right now I don't see anything in API responses that would let us figure that out. Probably this will be a postMVP endeavor.


src/pages/sighting/Annotations.jsx
```diff
@@ -12,6 +12,7 @@ import AnnotatedPhotograph from '../../components/AnnotatedPhotograph';
 import AnnotationEditor from '../../components/AnnotationEditor';
 import Text from '../../components/Text';
 import ConfirmDelete from '../../components/ConfirmDelete';
+import KickoffAnnotationDialog from './identification/KickoffAnnotationDialog';
```
Atticus29, line 15: This is actually kicking off ID rather than annotation, right? A little bit of a confusing name.

src/pages/sighting/encounters/Encounters.jsx
```diff
@@ -117,7 +117,7 @@ export default function Encounters({
   return (
     <div>
       <ConfirmDelete
-        open={encounterToDelete && !userDataLoading}
+        open={Boolean(encounterToDelete && !userDataLoading)}
```
Atticus29, line 120: Boolean not required here, right?

src/pages/sighting/identification/CommitDialog.jsx
```diff
@@ -0,0 +1,127 @@
...
+        <Button display="basic" onClick={onClose} id="CANCEL" />
+        <Button
+          loading={patchAgsLoading || commitAgsLoading}
+          display="primary"
+          onClick={async () => {
+            let idConfigPatchValue = [idConfig];
+            if (mode === jobModes.default) {
+              idConfigPatchValue = [
+                {
+                  algorithms: ['hotspotter_nosv'],
```
Atticus29, line 97: Seems weird to hard-code a default algo like this?

src/pages/sighting/identification/buildMatchingSetQuery.js
```diff
@@ -0,0 +1,58 @@
+import { flattenDeep, get } from 'lodash-es';
+
+function addChildMatches(rootRegion, matchId, insideRegion) {
+  const matchFound = rootRegion?.id === matchId;
+  const insideRegionNew = matchFound || insideRegion;
```
Atticus29, line 5: I'm  a little confused by this variable name. If a match is found at the current.... level, let's call it?.... why would insideRegionNew be true?

In other words, if it matches at root, why there be a new insideRegion?

src/pages/sighting/identification/buildMatchingSetQuery.js
```diff
@@ -0,0 +1,58 @@
...
+  const parentMatch = insideRegionNew ? [rootRegion] : [];
+  const children = rootRegion?.locationID || [];
```
Atticus29, line 7: Same here... these look like specific kinds of children (locationIdChildren?).

Atticus29: Oh! That's right... locationId is the name of recursive nodes in locationId.json. I'm guessing that this stems from that.

Atticus29: I guess maybe they could still have a more descriptive name than children. 

May I suggest a 500 character name 😈 ?

Atticus29: childMatches?

Atticus29 (changes requested):
A few things that I think might be worth changing, a few things I leave up to the author's discretion.

Also, perhaps it's worth noting that when a user wants to re-run identification, the CommitDialog is not utilized. This means that the end user is stuck with the exact parameters of their original ID job forever.

Not sure if this is within the scope of this ticket to accommodate.

src/pages/sighting/encounters/Encounters.jsx
brmscheiner: Actually this fixes a console warning. `null && null` is actually null, not false, so there was a material-ui warning that open was given an invalid type

src/pages/sighting/identification/CommitDialog.jsx
brmscheiner: yeah, eh, it's the mvp default. we will refine over time.

src/pages/sighting/identification/buildMatchingSetQuery.js
brmscheiner: would descendants be a better word? children does have another meaning in the context of react. matches doesn't make total sense to me because they aren't necessarily matches, just the children of the region that is actively being looked at

src/pages/sighting/identification/buildMatchingSetQuery.js
brmscheiner: so `addChildMatches` is called with `insideRegion` set to true if the matching region has already been found and `rootRegion` is a descendant of that region. if that's not true yet and a match is found, then `insideRegion` will be true at the next level for all of the current region's descendants. hence `insideRegionNew`. open to other names if you can think of something better

src/hooks/useMutate.js
```diff
@@ -9,9 +9,15 @@ const methods = {
...
+function formatError(failureObject) {
   try {
-    return response?.error ? response.error.toJSON().message : null;
+    const jsError = failureObject?.error;
+    if (jsError) return jsError.toJSON().message;
+    const houstonErrorMessage =
+      failureObject?.response?.data?.message;
+    if (houstonErrorMessage) return houstonErrorMessage;
+    const browserErrorMessage = failureObject?.message;
+    return browserErrorMessage || null;
```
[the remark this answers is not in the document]
brmscheiner: this function returned null or a string before, so i just kept it the same way to make sure i didnt mess with any other usages of it in the app

src/components/RadioChoice.jsx
```diff
@@ -0,0 +1,51 @@
...
+      <RadioGroup
+        aria-label="sighting-type"
+        name="sightingType"
+        value={value}
+        onChange={e => onChange(e.target.value)}
+      >
+        {choices.map(choice => {
+          const choiceLabel = choice?.labelId
+            ? intl.formatMessage({ id: choice.labelId })
+            : choice.label;
```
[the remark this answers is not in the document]
brmscheiner: we should be giving one or the other, otherwise this component can error

brmscheiner:
@Atticus29 
> Also, perhaps it's worth noting that when a user wants to re-run identification, the CommitDialog is not utilized. This means that the end user is stuck with the exact parameters of their original ID job forever.

this is very much worth noting and kinda a silly consequence of how this is currently implemented. BE folk are aware of the problem but we probably wont be seeing a new API for this any time soon

src/components/RadioChoice.jsx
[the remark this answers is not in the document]
Emily-Ke: If choice is definitely going to have either a labelId or a label, then the optional chaining operator is unnecessary.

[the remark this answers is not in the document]
brmscheiner: how do you mean? either labelId or label will be present, but not both. ha maybe i should do a proptypes

src/pages/sighting/identification/buildMatchingSetQuery.js
Atticus29: That makes more sense now. I'm ok with that name... Also ok with something like currentlyInsideRegion.

src/models/identification/useIdConfigSchemas.js
```diff
@@ -0,0 +1,91 @@
...
+      if (loading || error) return [];
+
+      const regionChoices = get(
+        data,
+        ['site.custom.regions', 'value', 'locationID'],
+        [],
+      );
+
+      const siteSpecies = get(data, ['site.species', 'value'], []);
+      const siteItisIds = siteSpecies.map(species => species.itisTsn);
```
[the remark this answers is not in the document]
brmscheiner: nah ITIS is a thing. https://www.itis.gov/. it's the international system for organizing taxonomies

Atticus29 (approved):
New changes LGTM.

PR #311, opened 2022-04-18
Dex 648 encounter prior to manual annotation sighting

Atticus29:


Atticus29:
Whoops. Forgot to provide a description.

Ok. This ticket:

pending vs. non-pending state now tracked to the level of AnnotationCreator component.

If pending, manual annotation just creates the usual orphan annotation... the end user has to attach this annotation to an animal before committing as per previous workflow.

New stuff: if not pending, manual annotation creates an encounter/animal first, tracks the new encounter ID created from this action, and creates an annotation using this encounter. The key here was handling an optional encounter ID parameter for the postAnnotation() call.

If that subsequent postAnnotation call was successful, I took the liberty of attaching this annotation to the animal with addAnnotationsToSightingEncounter().

If you create an annotation manually in a committed sighting, you should a) no longer see an error at attempting this and b) now see it attached to a new animal in the "Animals" tab.

Errors should be displayed one after another in the same alert, and the dialog should not immediately close unless there are no errors.

src/components/AnnotationCreator.jsx
```diff
@@ -19,27 +19,96 @@ import usePostAnnotation from '../models/annotation/usePostAnnotation';
...
+    sightingId,
+    copiedProperties,
+  );
+  const encounterGuidsAfterCreation = map(
+    get(
+      encounterCreationResponse,
+      ['response', 'data', 'encounters'],
+      [],
+    ),
+    encounter => encounter.guid,
```
brmscheiner, line 53: `encounter?.guid` here please

src/components/AnnotationCreator.jsx
```diff
@@ -229,16 +301,49 @@ export default function AnnotationCreator({
...
+                if (addToSightingEncounterError) {
+                  setErrors(prevErrors => [
+                    ...prevErrors,
+                    addToSightingEncounterError,
+                  ]);
+                }
+              }
               refreshSightingData();
-              onClose();
+              closeIfNoErrors();
```
brmscheiner, line 346: instead of doing this, what about looking at the response from `await addAnnotationsToSightingEncounter` and testing whether `response?.status === 200`?

src/components/AnnotationCreator.jsx
```diff
@@ -229,16 +301,49 @@ export default function AnnotationCreator({
...
+              newEncounterGuid,
             );
+            if (postAnnotationError)
+              setErrors(prevErrors => [
+                ...prevErrors,
+                postAnnotationError,
+              ]);
+
             if (newAnnotationId) {
+              if (newEncounterGuid) {
```
brmscheiner, line 333: simplify with `if (newAnnotationId && newEncounterGuid)`

src/components/AnnotationCreator.jsx
```diff
@@ -229,16 +301,49 @@ export default function AnnotationCreator({
               ),
             ];
             const theta = get(rect, 'theta', 0);
+            const newEncounterGuids = pending
+              ? []
+              : await createEncounter(
```
brmscheiner, line 306: so, instead of doing createEncounter and then postAnnotation, it's very important that we only continue moving through this process if the first request worked correctly. otherwise we could easily wind up with random new annotations that shouldn't be there.

plus, if you wait for success before moving on to the next step, you should never have multiple errors, only one error at a time. so it should simplify your life to not have to deal with an array of errors

src/models/annotation/usePostAnnotation.js
```diff
@@ -14,8 +14,12 @@ export default function usePostAnnotation() {
     rect,
     viewpoint,
     theta = 0,
+    encounterGuid = null,
   ) => {
     try {
+      const additionalProperties = encounterGuid
```
brmscheiner, line 20: we should be migrating request hooks to `useMutate` instead of continuing to develop/modify with the existing hooks. if you can use an old one as-is that's okay i guess, but at the point where we are changing/enhancing them we really should be migrating

src/models/encounter/useAddEncounter.js
```diff
@@ -29,17 +29,17 @@ export default function useAddEncounter() {
...
       setError(formatError(patchResponse));
       setSuccess(false);
-      return false;
+      return { success: false, response: formatError(patchResponse) };
     } catch (postError) {
       setLoading(false);
       setError(formatError(postError));
       setSuccess(false);
-      return false;
+      return { success: false, response: postError };
```
brmscheiner, line 42: same thing here.

src/pages/sighting/encounters/Encounters.jsx
```diff
@@ -352,7 +352,7 @@ export default function Encounters({
             'timeSpecificity',
             'locationId',
           ]);
-          await addEncounter(sightingId, copiedProperties);
+          await addEncounter(sightingId, copiedProperties); // Do we really want to proceed as usual if there's an error here?
```
brmscheiner, line 355: once we migrate to `useMutate` i think we can remove the refresh function below. in that implementation data should only be updated if the mutation is successful

brmscheiner (changes requested):

src/components/AnnotationCreator.jsx
```diff
@@ -19,27 +19,96 @@ import usePostAnnotation from '../models/annotation/usePostAnnotation';
...
   onClose,
   sightingData,
   refreshSightingData,
+  pending,
 }) {
   const [viewpoint, setViewpoint] = useState('');
   const [IAClass, setIAClass] = useState('');
   const [rect, setRect] = useState({});
   const [anchorEl, setAnchorEl] = useState(null);
+  const [errors, setErrors] = useState([]);
```
Emily-Ke, line 79: Because you are only using this to store other persisted values, it shouldn't be necessary.
`errors` can just be calculated from data that you already have.

src/components/AnnotationCreator.jsx
Atticus29: Done.

But also, if it's in a map, wouldn't it only ever be run if encounter existed?

src/components/AnnotationCreator.jsx
Atticus29: newAnnotationId gets created both if pending and if !pending.
However, addAnnotationsToSightingEncounter should only be run if !pending (and, indeed, more specifically only if there is a newEncounterGuid that can be used for said addAnnotationsToSightingEncounter call).

In other words, the two if statements above really ought to remain separate.

src/components/AnnotationCreator.jsx
Atticus29: If pending is true, addAnnotationsToSightingEncounter ought never get called. So, using it's response as the deciding factor of whether the dialog should be closed would not work in the pending === true case.

src/components/AnnotationCreator.jsx
Atticus29: Because pending sightings (i.e., AGSs) ought to be able to create orphan annotations, we need to allow orphan annotation creation to be possible.

Your point about only one error at a time is well-taken.

src/models/annotation/usePostAnnotation.js
Atticus29: Done.

src/models/encounter/useAddEncounter.js
Atticus29: Done

src/components/AnnotationCreator.jsx
```diff
@@ -19,27 +19,91 @@ import usePostAnnotation from '../models/annotation/usePostAnnotation';
...
+    addEncounterToSightingLoading ||
+    addAnnotationToSightingEncounterLoading ||
+    postAnnotationLoading;
+  const hasErrors =
+    addEncounterToSightingError ||
+    addAnnotationsToSightingEncounterError ||
+    postAnnotationError;
+  const onCancel = () => {
+    onClose();
+  };
```
Emily-Ke, line 106: Do the mutation errors need to be cleared here?

src/components/AnnotationCreator.jsx
```diff
@@ -188,26 +251,43 @@ export default function AnnotationCreator({
...
-          {error}
-        </CustomAlert>
+          {addEncounterToSightingError && (
+            <Text key={addEncounterToSightingError} variant="body2">
+              {addEncounterToSightingError}
+            </Text>
+          )}
+          {addAnnotationsToSightingEncounterError && (
+            <Text
+              key={addAnnotationsToSightingEncounterError}
```
brmscheiner, line 272: keys shouldn't be necessary unless you are iterating. were you getting a key error?

brmscheiner (approved):

src/components/AnnotationCreator.jsx
Atticus29: Apparently not! They were gone when I either cancelled or dismissed the dialog. I was similarly perplexed!

PR #314, opened 2022-04-21
DEX 845 - cleanup encounters table wrapping

Emily-Ke:
This prevents the individual's sightings table cells from wrapping. When a cell's text overflows and a user hovers over the cell, a tooltip appears with the full cell value. For cells that have not overflowed, there is no tooltip.
encounters-card-after

I updated all of the cell renderers instead of just the ones used by the EncountersCard in order to ensure that this design would work for them.

Hiding overflow required a sacrifice of the automatic table layout. For this table, I chose to let the Date column take a percentage of the space, let the Actions column take a fixed, calculated amount of space, and let the Location and Owner columns split the rest.

The actions cells do not have a `white-space: nowrap` style applied. However, they should not wrap because the calculation mentioned above should always provide enough width.

### DIFF
--- a/src/components/Button.jsx
+++ b/src/components/Button.jsx
@@ -1,19 +1,22 @@
-import React from 'react';
+import React, { forwardRef } from 'react';
 import { FormattedMessage } from 'react-intl';
 import { useTheme } from '@material-ui/core/styles';
 import Button from '@material-ui/core/Button';
 import CircularProgress from '@material-ui/core/CircularProgress';
 import BackIcon from '@material-ui/icons/KeyboardBackspace';
 
-const Core = function({
-  children,
-  display = 'panel',
-  loading = false,
-  style,
-  disabled,
-  size,
-  ...rest
-}) {
+const Core = function(
+  {
+    children,
+    display = 'panel',
+    loading = false,
+    style,
+    disabled,
+    size,
+    ...rest
+  },
+  ref,
+) {
   const theme = useTheme();
 
   let variant = undefined; // eslint-disable-line
@@ -33,6 +36,7 @@ const Core = function({
           padding: '4px 16px',
           ...style,
         }}
+        ref={ref}
         {...rest}
       >
         {children}
@@ -96,6 +100,7 @@ const Core = function({
       <button
         type="button"
         style={{ ...roleStyles, ...style }}
+        ref={ref}
         {...rest}
       >
         {loading ? (
@@ -119,6 +124,7 @@ const Core = function({
       disabled={disabled}
       style={{ ...roleStyles, ...style }}
       size={size}
+      ref={ref}
       {...rest}
     >
       {loading ? (
@@ -130,22 +136,23 @@ const Core = function({
   );
 };
 
-export default function CustomButton({
-  id,
-  domId = undefined,
-  values,
-  children,
-  ...rest
-}) {
+const CoreForwardRef = forwardRef(Core);
+
+function CustomButton(
+  { id, domId = undefined, values, children, ...rest },
+  ref,
+) {
   if (!id)
     return (
-      <Core id={domId} {...rest}>
+      <CoreForwardRef id={domId} ref={ref} {...rest}>
         {children}
-      </Core>
+      </CoreForwardRef>
     );
   return (
-    <Core id={domId} {...rest}>
+    <CoreForwardRef id={domId} ref={ref} {...rest}>
       <FormattedMessage id={id} values={values} />
-    </Core>
+    </CoreForwardRef>
   );
 }
+
+export default forwardRef(CustomButton);
--- a/src/components/cards/EncountersCard.jsx
+++ b/src/components/cards/EncountersCard.jsx
@@ -58,31 +58,56 @@ export default function EncountersCard({
       reference: 'date',
       name: 'time',
       label: 'Date',
-      options: { cellRenderer: cellRendererTypes.specifiedTime },
+      options: {
+        cellRenderer: cellRendererTypes.specifiedTime,
+        cellRendererProps: { noWrap: true },
+        width: '20%',
+      },
     },
     {
       reference: 'location',
       name: 'formattedLocation',
       label: 'Location',
-      options: { cellRenderer: cellRendererTypes.location },
+      options: {
+        cellRenderer: cellRendererTypes.location,
+        cellRendererProps: { noWrap: true },
+      },
     },
     {
       reference: 'owner',
       name: 'owner',
       label: 'Owner',
-      options: { cellRenderer: cellRendererTypes.user },
+      options: {
+        cellRenderer: cellRendererTypes.user,
+        cellRendererProps: { noWrap: true },
+      },
     },
     {
       reference: 'actions',
       name: 'guid',
       label: 'Actions',
       options: {
+        width: (function buildWidth() {
+          const numberOfActions = 2;
+          // the following values were determined by inspecting the resulting DOM elements.
+          const actionWidth = '1.5rem';
+          const totalActionPadding = `${numberOfActions *
+            theme.spacing(1)}px`;
+          const tableCellPadding = `${theme.spacing(4)}px`;
+          const actionLabelWidth = '73px';
+
+          const actionTdWidth = `calc(${numberOfActions} * ${actionWidth} + ${totalActionPadding} + ${tableCellPadding})`;
+          const actionThWidth = `calc(${actionLabelWidth} + ${tableCellPadding})`;
+          return `max(${actionTdWidth}, ${actionThWidth})`;
+        })(),
         customBodyRender: (_, encounter) => [
           <ActionIcon
+            key="view"
             variant="view"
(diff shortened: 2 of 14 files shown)


src/components/dataDisplays/DataDisplay.jsx
```diff
@@ -257,10 +258,22 @@ export default function DataDisplay({
...
+              <col
+                key={c.name}
+                style={
+                  get(c, 'options.width')
+                    ? { width: c.options.width }
+                    : {}
+                }
+              />
+            ))}
+          </colgroup>
```
Emily-Ke, line 276: This will create `colgroup` and `col` tags for every DataDisplay component. Because they are empty they do not affect the table's appearance.

**NOTE:** this logic will need to be updated if a table with an "expand" row column wants to set custom column widths.

src/components/cards/EncountersCard.jsx
```diff
@@ -58,31 +58,56 @@ export default function EncountersCard({
...
+        cellRenderer: cellRendererTypes.user,
+        cellRendererProps: { noWrap: true },
+      },
     },
     {
       reference: 'actions',
       name: 'guid',
       label: 'Actions',
       options: {
+        width: (function buildWidth() {
```
brmscheiner, line 90: this is cool, feel like it could benefit from abstraction though since we would probably want this set on all action columns in all tables. a few ideas: 
 - build an actions cellRenderer and accept a `numberOfActions` prop for this renderer
 - create a hook that returns the `buildActionColumnWidth` function 
 - something else?

src/components/dataDisplays/cellRenderers/DateRenderer.jsx
```diff
@@ -1,15 +1,34 @@
...
   accessor = 'created',
+  noWrap = false,
 }) {
-  return (
-    <Text component="span" variant="body2">
-      {formatDate(get(datum, accessor, ''))}
-    </Text>
+  const formattedValue = formatDate(get(datum, accessor, ''));
+
+  const coreComponent = <CoreForwardRef value={formattedValue} />;
```
brmscheiner, line 25: not really important but whenever I set a component to a variable I always capitalize that variable's name (ie CoreComponent)

src/components/dataDisplays/cellRenderers/OverflowController.jsx
```diff
@@ -0,0 +1,61 @@
...
+    () => {
+      if (!refEl) return;
+
+      const debouncedHandleResize = debounce(function handleResize() {
+        const { clientWidth, scrollWidth } = refEl;
+        const overflow = scrollWidth - clientWidth > 0;
+        setIsOverflowed(overflow);
+      }, 200);
+
+      const resizeObserver = new ResizeObserver(
```
brmscheiner, line 31: if we encounter performance problems with this logic on large tables, we might want to consider a solution that involves one resize observer per column instead of per cell

src/components/dataDisplays/cellRenderers/OverflowController.jsx
```diff
@@ -0,0 +1,61 @@
...
+      };
+    },
+    [refEl],
+  );
+
+  const childrenWithStyles = React.cloneElement(
+    React.Children.only(children),
+    {
+      ref,
+      style: noWrapStyles,
```
brmscheiner, line 47: would this override styles being passed to the children? if so we might need to merge noWrapStyles with any other styles

src/components/cards/EncountersCard.jsx
Emily-Ke: Created a hook in https://github.com/WildMeOrg/codex-frontend/pull/314/commits/9360ac70598a7a71a19de6dd510d94321b772ecc

src/components/dataDisplays/cellRenderers/DateRenderer.jsx
Emily-Ke: Fixed in https://github.com/WildMeOrg/codex-frontend/pull/314/commits/9ef13bfed1e6b2cff6894f501e1c2dee3cb9a077

src/components/dataDisplays/cellRenderers/OverflowController.jsx
Emily-Ke: Yes! Also, if a ref is passed to the child, this new ref is ignored. Currently none of the cell renderers accept style props or refs though.

Atticus29 (approved):
A whole lot of this was very new to me. I especially had trouble following OverflowController.jsx. 

I didn't see anything that looked weird to me, but please take that with a huge grain of salt. 

Lots of questions that are in no way urgent or need to be followed up on before May 1.

src/components/Button.jsx
```diff
@@ -130,22 +136,23 @@ const Core = function({
...
 };
 
-export default function CustomButton({
-  id,
-  domId = undefined,
-  values,
-  children,
-  ...rest
-}) {
+const CoreForwardRef = forwardRef(Core);
```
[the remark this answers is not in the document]
Atticus29: As a follow up... I never found the place in the code where the reference was actually needed/used.

src/components/dataDisplays/cellRenderers/OverflowController.jsx
```diff
@@ -0,0 +1,61 @@
...
+
+  const ref = useCallback(node => {
+    setRefEl(node);
+  }, []);
+
+  useEffect(
+    () => {
+      if (!refEl) return;
+
+      const debouncedHandleResize = debounce(function handleResize() {
```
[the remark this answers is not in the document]
Emily-Ke: It will prevent the resize handler from running as often.

src/components/dataDisplays/cellRenderers/OverflowController.jsx
```diff
@@ -0,0 +1,61 @@
...
+  whiteSpace: 'nowrap',
+  overflow: 'hidden',
+  textOverflow: 'ellipsis',
+};
+
+export default function OverflowController({ title, children }) {
+  const [isOverflowed, setIsOverflowed] = useState(false);
+  const [refEl, setRefEl] = useState();
+
+  const ref = useCallback(node => {
```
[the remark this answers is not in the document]
Emily-Ke: 1. From the React docs: ([useCallback](https://reactjs.org/docs/hooks-reference.html#usecallback)):
> `useCallback(fn, deps)` is equivalent to `useMemo(() => fn, deps)`.
2. It shouldn't be called continuously because of the way that refs are handled.
> React will call the ref callback with the DOM element when the component mounts, and call it with null when it unmounts. ([callback refs](https://reactjs.org/docs/refs-and-the-dom.html#callback-refs))

src/components/dataDisplays/cellRenderers/CapitalizedStringRenderer.jsx
```diff
@@ -1,9 +1,32 @@
...
+  value,
+  noWrap = false,
+}) {
+  const capitalizedValue = capitalize(value);
+
+  const CoreComponent = <CoreForwardRef value={capitalizedValue} />;
+
+  return noWrap ? (
+    <OverflowController title={capitalizedValue}>
+      {CoreComponent}
```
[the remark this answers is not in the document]
Emily-Ke: `capitalizedValue` is not a component, it is a string. When there is overflow, the string will end up in two places: Tooltip adds it as a title attribute on the underlying DOM element and CoreComponent adds it as a text node to the same element.
For example, if I pass 'hello' as the value prop to CapitalizedStringRender, capitalizedValue will be 'Hello' and, simplified, I should end up with:
```
<p title="Hello">Hello</p>
```

src/components/dataDisplays/DataDisplay.jsx
Emily-Ke: It is allowing me to set custom width values for columns in the table. Without it, the columns would each take approximately the same amount of horizontal space. It looks really strange when one column has clipped text and there is a large section of white space before the next column's content.

I do think that this logic will probably undergo some changes if or as we apply noWrap to other tables.

src/components/Button.jsx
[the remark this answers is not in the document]
Emily-Ke: 1. Correct.
2. LocationRenderer conditionally renderers Button. If you mean more generally, the refs are used by both OverflowController and Tooltip.

src/components/dataDisplays/cellRenderers/OverflowController.jsx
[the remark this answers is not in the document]
brmscheiner: events like `onPageResize` or `onScroll` fire continuously as a user is performing the action (ie. as long as their mouse is depressed after they clicked on the edge of the browser). it depends on the users system, but they fire again and again as fast as possible, you could get 50 events in a second

src/components/dataDisplays/cellRenderers/CapitalizedStringRenderer.jsx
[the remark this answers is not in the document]
Atticus29: Makes sense now! Thanks!

PR #317, opened 2022-04-22
Bulk import: validate assets on spreadsheet submission

brmscheiner:
image


Emily-Ke (changes requested):
`validateAssetStrings` returns an error message that "The following asset(s) were not found and will be ignored: ." if there are no assets.

PR #319, opened 2022-04-25
Remove fake cooccurrences table

brmscheiner:
Doesn't look like this API will exist in time for MVP, so we are just going to comment out the table for now.


Atticus29 (approved):
LGTM

colinwkingen:
For the record the API exists and is tested. I think we aren't going to spend the time to connect these dots this week though.

Emily-Ke:
We should be able to delete the `fakeCoocs.js` file too then.

brmscheiner:
@Emily-Ke I'm leaving the fake data for now so that when we uncomment it we can still see what we had before. I think it will be easier to develop that way.

PR #320, opened 2022-04-26
DEX-939 restore missing bulk import fields

Emily-Ke:
# DEX-939
- Uses `assetReferences` as the key for assets in the bulk import workflow, which restores "Assets" to the available sighting fields list.
- Uses `firstName` as the key for individual names in the bulk import workflow, which restores "Individual Name" to the available animal fields list. I checked with BE and the key `firstName` on the `encounter` is the contract that is expected.
- Adds the sighting `comments` key to the bulkImportCategories sighting fields so that "Notes" correctly appears in the available sighting fields lists. Because it is possible for sighting comments to conflict in the spreadsheet entries, only the last comment that has a value for a given sighting is persisted to the sighting.

# Additional changes
- The last `time` and `timeSpecificity` for a given sighting are saved to that sighting instead of the first, in order to be consistent with the other fields.
- I changed the message that displays after a successful Flatfile submission from "{count} sightings imported." to "{count} encounters imported." to more accurately reflect what was imported.

### DIFF
--- a/src/pages/bulkImport/constants/bulkReportConstants.js
+++ b/src/pages/bulkImport/constants/bulkReportConstants.js
@@ -5,7 +5,12 @@ export const bulkImportCategories = {
   },
   sighting: {
     name: 'sighting',
-    fields: ['sightingId', 'assetReferences', 'locationId'],
+    fields: [
+      'sightingId',
+      'assetReferences',
+      'locationId',
+      'comments',
+    ],
   },
   shared: {
     name: 'shared',
--- a/src/pages/bulkImport/constants/bulkReportConstants.js
+++ b/src/pages/bulkImport/constants/bulkReportConstants.js
@@ -5,7 +5,7 @@ export const bulkImportCategories = {
   },
   sighting: {
     name: 'sighting',
-    fields: ['sightingId', 'assets', 'locationId'],
+    fields: ['sightingId', 'assetReferences', 'locationId'],
   },
   shared: {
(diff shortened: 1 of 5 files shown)


brmscheiner (approved):
works great!

src/pages/bulkImport/constants/bulkReportConstants.js
```diff
@@ -50,5 +55,5 @@ export const bulkFieldSchemas = [
   { name: 'timeMinutes', labelId: 'TIME_MINUTES' },
   { name: 'timeSeconds', labelId: 'TIME_SECONDS' },
   { name: 'utcOffset', labelId: 'TIMEZONE' },
-  { name: 'assets', labelId: 'ASSETS' },
+  { name: 'assetReferences', labelId: 'ASSETS' },
```
Atticus29, line 58: What about comments?

Atticus29 (approved):
LGTM!

I'm noticing comments missing from the available fields list, but seems very minor.
Screen Shot 2022-04-26 at 7 06 12 PM

Emily-Ke:
`comments` is the key for the label "Notes", which is listed under sighting fields.
The key/label relationship is defined in `useSightingFieldSchemas.js `:https://github.com/WildMeOrg/codex-frontend/blob/0415bb80479feec08081d32a87cce993e064b84e/src/models/sighting/useSightingFieldSchemas.js#L149

PR #325, opened 2022-04-28
Sighting search updates

brmscheiner:
image

image

Sorry for the fluff here, I forgot to run npm format in a previous commit

### DIFF
--- a/locale/en.json
+++ b/locale/en.json
@@ -523,6 +523,8 @@
   "BASIC_INPUTS": "BASIC FIELDS",
   "COMPOUND_INPUTS": "COMPOUND FIELDS",
   "SPECIAL_INPUTS": "SPECIAL FIELDS",
+  "SIGHTING_SEARCH_RESULT_PRIMARY_TEXT": "{date} sighting in {region}",
+  "SIGHTING_SEARCH_RESULT_SECONDARY_TEXT": "Reported by {name} on {date}",
   "SELECTED_QUERY_ANNOTATION": "Selected query annotation",
   "IDENTIFICATION_FINISHED_TIME": "Identification finished on {time}.",
   "SELECTED_MATCH_CANDIDATE": "Selected match candidate",
--- a/locale/en.json
+++ b/locale/en.json
@@ -637,6 +637,7 @@
   "DELETE_ANNOTATION_CONFIRMATION": "Click save to delete this annotation.",
   "DELETE_THIS_ANNOTATION": "Delete this annotation",
   "SUBMITTED_BY": "Submitted by ",
+  "REPORTER": "Reporter",
   "PROJECT_NAME": "Project name",
   "INDIVIDUAL_NOT_FOUND": "Individual not found",
   "START_DATE": "Start date",
--- a/src/components/AuthenticatedAppHeader/SightingsButton.jsx
+++ b/src/components/AuthenticatedAppHeader/SightingsButton.jsx
@@ -1,8 +1,11 @@
 import React, { useState } from 'react';
+import { FormattedMessage } from 'react-intl';
+import { get } from 'lodash-es';
 
 import List from '@material-ui/core/List';
 import Divider from '@material-ui/core/Divider';
 
+import { formatDate } from '../../utils/formatters';
 import useSightingTermQuery from '../../models/sighting/useSightingTermQuery';
 import Text from '../Text';
 import SearchButton from './SearchButton';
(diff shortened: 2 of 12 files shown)


src/components/dataDisplays/SightingsDisplay.jsx
```diff
@@ -32,58 +30,56 @@ export default function SightingsDisplay({ sightings, loading }) {
...
       options: {
-        cellRenderer: cellRendererTypes.date,
-        cellRendererProps: { accessor: 'lastSeen' },
+        customBodyRender: owners => {
+          const ownerName = get(
+            owners,
+            [0, 'full_name'],
+            'Unknown User',
+          );
+          const ownerGuid = get(owners, [0, 'guid'], 'Unknown User');
```
Emily-Ke, line 55: Do you want a potential link to `/users/Unknown%20User`? It would be cleaner to not return a link for a user without a guid.

Emily-Ke (approved):

src/components/AuthenticatedAppHeader/SightingsButton.jsx
```diff
@@ -58,18 +64,40 @@ export default function SightingsButton() {
...
+                      values={{
+                        region: regionLabel,
+                        date: sightingDate,
+                      }}
+                      id="SIGHTING_SEARCH_RESULT_PRIMARY_TEXT"
+                    />
+                  }
+                  secondaryText={
+                    <FormattedMessage
+                      values={{ name: ownerName, date: createdDate }}
```
Atticus29, line 97: createdDate has fancy formatting, while specified time has yyyy-MM-dd formatting.

src/components/dataDisplays/SightingsDisplay.jsx
```diff
@@ -32,58 +30,56 @@ export default function SightingsDisplay({ sightings, loading }) {
...
-                {i !== 0 && ', '}
-                <Link href={`/individuals/${id}`}>{id}</Link>
-              </span>
-            ))}
-          </span>
-        ),
-      },
+      name: 'locationId_value',
+      labelId: 'REGION',
+      align: 'left',
```
Atticus29, line 42: There's a locationRenderer; don't know if you are intentionally avoiding it here.

src/components/dataDisplays/SightingsDisplay.jsx
```diff
@@ -32,58 +30,56 @@ export default function SightingsDisplay({ sightings, loading }) {
...
-      label: intl.formatMessage({ id: 'SUBMITTED_BY' }),
+      name: 'guid',
+      labelId: 'ACTIONS',
       options: {
-        customBodyRender: value => (
-          <Link href={`/users/${value}`}>{value}</Link>
-        ),
+        customBodyRender: guid => {
+          return (
+            <ActionIcon
```
Atticus29, line 74: Probably makes more sense for this to be newTab

Atticus29 (changes requested):
Screen Shot 2022-04-28 at 8 25 22 PM
I'm not seeing the regions dropdown populating.

locale/en.json
```diff
@@ -523,6 +523,8 @@
   "BASIC_INPUTS": "BASIC FIELDS",
   "COMPOUND_INPUTS": "COMPOUND FIELDS",
   "SPECIAL_INPUTS": "SPECIAL FIELDS",
+  "SIGHTING_SEARCH_RESULT_PRIMARY_TEXT": "{date} sighting in {region}",
```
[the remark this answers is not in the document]
brmscheiner: that might sound better in some cases but it doesn't make sense grammatically in all cases. because this is a specified date it can take many forms, such as `2004`, `May 2004`, `May 2, 2004`, or `May 2, 2004 at 8:15`. So we would have "sighting in mozambique on 2004" for example...

src/components/AuthenticatedAppHeader/SightingsButton.jsx
brmscheiner: I think that's fine. Maybe later we can introduce fancy formatting for specifiedTimes

src/components/dataDisplays/SightingsDisplay.jsx
brmscheiner: Yeah, locationRenderer requires more resources to derive the region label from the region guid. Since the backend is just sending us the label no reason to convert the guid

src/components/dataDisplays/SightingsDisplay.jsx
brmscheiner: Ooops copypaste mistake!

Atticus29 (commented):
I'm still not seeing the region select menu generate anything

Atticus29 (approved):

PR #328, opened 2022-04-29
DEX-945 update Flatfile props

Emily-Ke:
- Moves FlatfileButton `devMode` and `managed` props into the settings prop so that they work correctly.
  - `devMode` causes a banner to display at the top of the dialog.
  - `managed` adds additional file options. When managed is false, the file options are limited to .csv, .tsv, and .txt spreadsheets
- Sets devMode to true or false conditionally based on the `__DEV__` global variable.
- Removes the `maxRecords` prop instead of moving it to settings. Currently it is unknown how many encounters houston can handle, and the bulk import processing implementation is changing.

after

### DIFF
--- a/src/pages/bulkImport/BulkReportForm.jsx
+++ b/src/pages/bulkImport/BulkReportForm.jsx
@@ -139,12 +139,11 @@ export default function BulkReportForm({ assetReferences }) {
           }}
         >
           <FlatfileButton
-            devMode
-            managed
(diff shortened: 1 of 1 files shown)


brmscheiner (approved):

Atticus29 (approved):
"Approved, and so it shall be" https://youtu.be/h_DHJrswOKw

PR #333, opened 2022-05-03
Dex 663 user manager revoke notifications

Atticus29:
Screen Shot 2022-05-03 at 4 43 56 PM

Really, the main change should be that I deleted useRevokeCollaboration (good riddance) because we can now use  usePatchCollaboration instead.


src/pages/individual/Individual.jsx
```diff
@@ -180,6 +180,8 @@ export default function Individual() {
     setError: setPatchError,
   } = usePatchIndividual();
 
+  const error = deleteError || patchError;
```
brmscheiner, line 183: as discussed in slack, i think this should be removed

src/pages/userManagement/UserManagerCollaborationEditTable.jsx
```diff
@@ -20,26 +21,40 @@ export default function UserManagersCollaborationEditTable({
   collaborationLoading,
   collaborationError,
 }) {
+  const {
+    data: currentUserData,
+    isFetching: userDataFetching,
```
brmscheiner, line 26: why fetching for this one and loading for the other one?

src/pages/userManagement/UserManagerCollaborationEditTable.jsx
```diff
@@ -170,7 +185,7 @@ export default function UserManagersCollaborationEditTable({
           style={{ margin: '8px 16px', display: 'block' }}
         />
       ) : null}
-      {isError && !dismissed ? (
+      {patchError && !dismissed ? (
```
brmscheiner, line 188: why keep track of dismissed instead of using the `onClearError` functionality we have built?

brmscheiner (changes requested):
Some quick changes to make before merge

src/pages/individual/Individual.jsx
Atticus29: Resolved by merge with develop

src/pages/userManagement/UserManagerCollaborationEditTable.jsx
Atticus29: No reason. I was just keeping a pattern that was already there. Changed to loading to keep in line with most everything else. Done.

src/pages/userManagement/UserManagerCollaborationEditTable.jsx
Atticus29: This code pre-dated our onClearError pattern. Fixed now.

brmscheiner (approved):

PR #334, opened 2022-05-04
Dex 975 fill animal1 with manual annotation if none

Atticus29:
Fixes the issue where if an annotation-less encounter already exists, a new encounter was being created anyway with the new manual annotation assigned to the latter.

With the fix, a new encounter is not created if an annotation-less encounter already exists, and the new manual annotation is assigned to the already-existing one instead. Only in the !pending case, of course.

### DIFF
--- a/src/components/AnnotationCreator.jsx
+++ b/src/components/AnnotationCreator.jsx
@@ -75,6 +75,8 @@ export default function AnnotationCreator({
   sightingData,
   pending,
 }) {
+  console.log('deleteMe sightingData is: ');
+  console.log(sightingData);
   const [viewpoint, setViewpoint] = useState('');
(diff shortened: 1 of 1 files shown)


src/components/AnnotationCreator.jsx
```diff
@@ -40,6 +43,13 @@ async function createEncounter(sightingData, addEncounterToSighting) {
     encounter => encounter?.guid,
     [],
   );
+  const encountersWithNoAnnotations = filter(
+    get(sightingData, 'encounters'),
```
brmscheiner, line 47: fall back to an empty array or you will get an error if sightingData.encounters does not exist

brmscheiner (approved):

PR #335, opened 2022-05-04
Image Upload and Tus Workflow Hardening

Emily-Ke:
These are the front-end changes that accompany the houston changes by @bluemellophone in wildmeorg/houston#631

- Adds an alert to the bulk import page that displays the current progress state of an asset group.
  - Shows the asset count as "pending" in the asset group table while the asset group sightings are in the preparation stage.
  - If the preparation progress is not complete, the page stays in a loading state for up to eight seconds in order to try to prevent a flash of the alert for progress that completes quickly.
  - *note*: I was having trouble with `useFetch`, so I decided to use useQuery directly for the new progress queries. I also switched to useQuery for fetching the asset groups. I named the files `use{Model}Query` to indicate that the API is different than for the other query hooks.
- Sends a delete request when a file is removed from the Uppy dashboard on the Report a Sighting or Bulk Import page.
  - Updates the `uuid` package in order to get access to the new `validate` function
  - If there is an error with the deletion, an error message displays on the dashboard. Unfortunately Uppy errors don't seem to have an option for remaining on screen until the user dismisses them.
- Removes the fingerprint that Tus stores in local storage after a successful upload. Because of the way that we assign resource IDs, there is currently no way that the local resource would be used again. This prevents the HEAD requests that the Tus uploader was making and prevents the resulting 404 errors from displaying in the console.
- Updates the code to use the new AGS schema introduced in the houston changes.
- Updates the eslint rules to allow the style prop on FormattedNumber components.
- I had also done some work on useOptions, but https://github.com/WildMeOrg/codex-frontend/commit/86cfb71a9fbc463384b0aa249b0525718fb42df0 made it obsolete. I kept the other clean up changes to `useOptions`, `useEncounterFieldSchemas`, and `useBulkImportFields` that resulted from my original effort.

Deletion error:
deletion-error

Progress on the bulk import page. This video was made with mock data, but the polling cadence is the same.

https://user-images.githubusercontent.com/50299119/168941400-26423403-3acb-4fdc-bde0-fa7be2a4bde3.mov

## TODO
- [ ] Add the asset preparation progress UI for the pending-sighting page


brmscheiner (commented):
What is the purpose of this? If I remove an image from uppy and then report a sighting, the report is still successful and the extra image does not show up.

src/models/progress/useProgressQuery.js
```diff
@@ -0,0 +1,28 @@
...
+import { isProgressSettled } from '../../utils/progressUtils';
+import {
+  getProgress,
+  queryOptions as defaultQueryOptions,
+} from './progressQueryDefaults';
+
+export default function useProgressQuery(progressGuid, queryOptions) {
+  const isSettledRef = useRef(false);
+
+  const result = useQuery(
```
brmscheiner, line 14: i'm not really understanding why you're using a homemade version of useQuery here instead of useFetch. i'm probably just missing it

src/models/progress/useProgressQuery.js
```diff
@@ -0,0 +1,28 @@
...
+
+import { getProgressQueryKey } from '../../constants/queryKeys';
+import { isProgressSettled } from '../../utils/progressUtils';
+import {
+  getProgress,
+  queryOptions as defaultQueryOptions,
+} from './progressQueryDefaults';
+
+export default function useProgressQuery(progressGuid, queryOptions) {
+  const isSettledRef = useRef(false);
```
brmscheiner, line 12: is useRef better than useState here? seems like maybe this is just a wilder version of `[isSettled, setIsSettled] = useState(false)`

brmscheiner: the one difference i can think of without looking at the docs is i dont think updating the ref will trigger a rerender. though in this case i think you'd actually want to rerender ASAP?

src/models/progress/useProgressStateQuery.js
```diff
@@ -0,0 +1,87 @@
...
+  const queryClient = useQueryClient();
+  const queryObserver = useMemo(
+    () => {
+      if (!progressGuid) return null;
+
+      const observerOptions = {
+        ...defaultQueryOptions,
+        queryKey: getProgressQueryKey(progressGuid),
+        queryFn: getProgress,
+        notifyOnPropsChange: ['data', 'error', 'isFetched'],
```
brmscheiner, line 27: again i'm curious what extra renders you were seeing

src/models/progress/useProgressStateQuery.js
```diff
@@ -0,0 +1,87 @@
...
+      };
+
+      if (enabled !== undefined) observerOptions.enabled = enabled;
```
brmscheiner, line 30: what if enabled were `false`? i'm confused about why you're only looking for undefined here. what about like `enabled: Boolean(enabled)`?

src/pages/assetGroup/AssetGroup.jsx
```diff
@@ -20,14 +27,59 @@ import MoreMenu from '../../components/MoreMenu';
...
+  );
+
+  const isProgressQueryEnabled = Boolean(
+    preparationProgressGuid && isAssetGroupPreparing,
+  );
+
+  const {
+    state: preparationProgressState,
+    error: preparationProgressError,
+  } = useProgressStateQuery(
```
brmscheiner, line 64: i'm really confused about the role this hook is playing. why not just `useProgressQuery` here and parse `data` with the utilities you have made to determine the status code and whether it's in a resolved state

src/models/assetGroup/useAssetGroupQuery.js
```diff
@@ -0,0 +1,29 @@
...
+    const response = await axios.get(apiUrl);
+    return response?.data;
+  } catch (error) {
+    const status = get(error, 'response.status');
+    throw { message: error?.message, statusCode: status };
+  }
+}
+
+export default function useAssetGroupQuery(assetGroupGuid) {
+  return useQuery(
```
brmscheiner, line 20: again why abandon useFetch here?

brmscheiner (changes requested):
there are some points where i was scratching my head. it's possible most of the code is good to go and i just need to understand why the normal pattern didn't work, but it's also possible the code could be simplified greatly with the same result.

src/models/progress/useProgressQuery.js
Emily-Ke: You are definitely right on this one. I'm going to try to find a native react-query way to update the options, but if I can't I am going to switch to useState.

src/models/progress/useProgressStateQuery.js
Emily-Ke: If I don't pass in enabled, I don't want to set that property on the options. That way it can use whatever the query's default value is.

src/pages/assetGroup/AssetGroup.jsx
Emily-Ke: If progress is pending, `useProgressQuery` will cause everything to re-render every two seconds. This hook triggers re-renders far less often. For AssetGroup, this may be overkill, I'm not sure, but for the sightings page, it will probably become more relevant.

src/models/progress/useProgressQuery.js
Emily-Ke: I was having trouble with `useFetch` being unreliable. It was just easier to use useQuery.

src/components/LinearProgressWithLabel.jsx
```diff
@@ -0,0 +1,39 @@
...
+}) {
+  const theme = useTheme();
+
+  return (
+    <Box display="flex" alignItems="center" style={containerStyle}>
+      <LinearProgress
+        {...linearProgressProps}
+        style={{
+          width: '100%',
+          marginRight: `${theme.spacing(1)}px`,
```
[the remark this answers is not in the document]
Emily-Ke: It looks like everywhere that it currently is used is code I have written. 😆 I don't really understand why you set up a Material UI theme to make that constraint easy to follow and then ignore it, but I'll switch it out for hard coded values.

src/components/assets/AssetsProcessingAlert.jsx
```diff
@@ -0,0 +1,172 @@
...
+export default function AssetsProcessingAlert({
+  progressGuid,
+  isAssetGroupPreparing,
+  ...rest
+}) {
+  const intl = useIntl();
+  const theme = useTheme();
+
+  const { isLoading, data, error } = useProgressQuery(progressGuid, {
+    notifyOnChangeProps: ['isLoading', 'data', 'error'],
```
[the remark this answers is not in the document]
Emily-Ke: yes

src/components/LinearProgressWithLabel.jsx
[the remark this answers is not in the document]
brmscheiner: ah, i'm not sure the codebase is entirely consistent about this, but where possible i'm trying to use theme for colors and not so much for spacing. we could migrate to using theme for spacing as well but before we do that i would want to make sure theme.spacing is set up to be on a 4px grid system (or maybe see if we want to switch over to the material defaults across the board)

src/models/progress/useProgressQuery.js
brmscheiner: nothing wrong with useState

src/models/progress/useProgressQuery.js
brmscheiner: if `useFetch` is unreliable, better to fix it! what was the problem exactly? 

don't want to abandon the patterns we are introducing if we can help it

src/pages/assetGroup/AssetGroup.jsx
brmscheiner: i think it might be premature optimization.. while we want to avoid unnecessary rerenders usually what we are worried about is something causing a ton of rerenders per second, one render every two seconds should be no big deal unless the component being rendered has a ton of issues

Emily-Ke:
This PR is getting split into smaller, more specific PRs.

PR #336, opened 2022-05-05
Iceland cleanup, switch over to new create individual UI

brmscheiner:


src/components/ButtonMenu.jsx
```diff
@@ -48,9 +49,11 @@ export default function ButtonMenu({
           <MenuItem
             key={action.id}
             onClick={e => {
-              action.onClick(e);
+              if (action.onClick) action.onClick(e);
               handleClose();
             }}
+            component={action.href ? RouterLink : undefined}
```
Atticus29, line 55: I don't understand why this is necessary.

Atticus29 (approved):
LGTM

src/components/ButtonMenu.jsx
brmscheiner: ahh, so that we are rendering a link instead of a button in the menu!

PR #338, opened 2022-05-09
Dex 934 twitter site settings

Atticus29:
Accommodates configuration of a twitterbot in such a way that other intelligent agents can easily be implemented by simply adding values to en.json and by adding a new object to codex-frontend/src/constants/intelligentAgentSchema.js. In other words, I tried to generalize this as much as I could.

You may notice that SettingsTextInput has a new skipDescription prop. This is because descriptions IDs come back from the intelligent agent back end definitions. Even if you set the corresponding Id (e.g., CONFIGURATION_INTELLIGENT_AGENT_TWITTERBOT_ENABLED_DESCRIPTION) in en.json to be '' or have it be missing entirely, the entire all-caps string was used as a default.

This PR also features refactors of several custom hooks: usePostSiteSettings, which saw usePostSiteSetting split into its own separate hook, and usePostSettingAsset.

Note that clearTwitterTestError in GeneralSettings is unused! I originally thought that onClearError, though written for useFetch, doesn't seem to be working and I couldn't figure out why. It had not been implemented anywhere else in the codebase. I believe that the problem is actually not that onClearError doesn't work; it works fine. Instead, useGetTwitterbotTestResults is getting called frequently and is correctly reporting what it thinks is an error (i.e., that twitterbot is disabled and/or misconfigured). So, I opted to make this a warning and make it non-dismissible.

You may also notice the showTwitterSuccess hook and wonder why I didn't use a setSuccess from useMutate. Well, it's because there is no setSuccess for useFetch (cf. useMutate), and I didn't think that there would be many other use cases to warrant the creation of one.

Finally, note that if twitterStatusCode is not also monitored as a dependency in the useEffect, the old failure doesn't get removed when valid credentials are used.

QA notes/expected behavior: 

When twitterbot is disabled, the user should not see the form fields for filliing out the twitter API credentials. They should be able to save their results and see a warning message that just lets them know that twitter is not configured. This should be non-dismissible.

When twitterbot is enabled, the user should see the the twitter API credential form fields. All credentials are required before the save option is enabled. 
Upon save, if the credentials are valid, the user should see an success alert that the site settings were changed successfully, but they should also separately see a information alert that their twitter configuration was successful (this should include some information about the twitterbot's handle and name). 

Upon save, if the credentials are invalid, the user should see an success alert that the site settings were changed successfully, but they should also separately see a warning that twitter is not enabled or configured successfully. This should be non-dismissible.

If there are values for the twitterbot api credentials already populating those fields and the user toggle the disable switch and saves, those value are removed before saving.

Screen Shot 2022-05-09 at 3 41 00 PM

Screen Shot 2022-05-09 at 3 41 09 PM
Screen Shot 2022-05-09 at 3 41 17 PM
Screen Shot 2022-05-09 at 3 41 26 PM
Screen Shot 2022-05-09 at 3 46 31 PM

A lot of things were refactored in this one: lots of potential for regressions. Test out adding/removing species, regions, relationship, customFields, configuring the splash settings... be sure to hard refresh to ensure that any changes have been persisted.


src/components/settings/SettingsTextInput.jsx
```diff
@@ -21,6 +21,7 @@ export default function SettingsTextInput({
   setCurrentValues,
   customFieldCategories,
   settingKey,
+  skipDescription = false,
```
brmscheiner, line 24: instead of having a prop here, could you just check whether `matchingSetting.description || matchingSetting.descriptionId` exists?

src/models/site/useGetTwitterbotTestResults.js
```diff
@@ -0,0 +1,12 @@
+import queryKeys from '../../constants/queryKeys';
+import useFetch from '../../hooks/useFetch';
+
+export default function useGetTwitterbotTestResults() {
+  return useFetch({
+    queryKey: queryKeys.twitterBot,
```
brmscheiner, line 6: where possible let's name query keys after the data that is returned. so not twitterBot here, but twitterBotTestResult

src/pages/generalSettings/IntelligentAgentSettings.jsx
```diff
@@ -0,0 +1,169 @@
...
+
+    const noCredsMissing =
+      intelligentAgentSettingsFields.filter(
+        currentField => get(currentValues, currentField) === '',
+      ).length === 0;
+
+    const allValid =
+      !isCurrentPlatformEnabled ||
+      (isCurrentPlatformEnabled && noCredsMissing);
+    setAllValid(allValid);
```
brmscheiner, line 44: if i'm reading correctly, this will cause setState to get called every time `IntelligentAgentSettings` is rendered. surprised there is no infinite loop here but either way this needs to be fixed

src/pages/generalSettings/IntelligentAgentSettings.jsx
```diff
@@ -0,0 +1,169 @@
...
+                  />
+                </Link>
+              ),
+            }}
+          />
+          {filter(
+            currentPlatformFields,
+            field =>
+              get(field, 'label') !== currentPlatformEnablingField,
+            [],
```
brmscheiner, line 107: this would read much clearer if you calculated this above and set it to a variable

brmscheiner (commented):
looks pretty good to me! you might want to add some QA notes to test other things that might be affected by the changes you made to the react-query stuff

brmscheiner:
a couple design comments

- if twitterbot is disabled we should not run the test or show the dialog 
- there should be 16px space between the various dialogs

Atticus29:
> 

Great minds. Already there!

Atticus29:
Note: this PR currently causes a regression with default fields patches.

src/models/site/useGetTwitterbotTestResults.js
Atticus29: Done

src/pages/generalSettings/IntelligentAgentSettings.jsx
Atticus29: Done

Atticus29:
> if twitterbot is disabled we should not run the test or show the dialog
> there should be 16px space between the various dialogs
> Note: this PR currently causes a regression with default fields patches.

Done.

src/pages/generalSettings/IntelligentAgentSettings.jsx
```diff
@@ -0,0 +1,187 @@
...
+              currentField => get(currentValues, currentField) === '',
+            ).length === 0;
+          return (
+            !isCurrentPlatformEnabled ||
+            (isCurrentPlatformEnabled && noCredsMissing)
+          );
+        },
+      );
+      const allSettingsValide = every(agentSettingsValid, Boolean);
+      setAllValid(allSettingsValide);
```
brmscheiner, line 43: typo here

src/pages/generalSettings/GeneralSettings.jsx
```diff
@@ -238,12 +320,16 @@ export default function GeneralSettings() {
...
+              putSiteSettings({ data: currentValues });
+              if (logoPostData)
+                postSettingsAsset({
+                  data: logoPostData,
+                });
             }}
             style={{ marginTop: 12 }}
             display="primary"
             loading={loading}
+            disabled={!allValid}
```
brmscheiner, line 332: is `allValid` a good name? seems like this variable is more about whether intelligentAgentFields are valid

brmscheiner (approved):
two small things

src/pages/generalSettings/IntelligentAgentSettings.jsx
Atticus29: Done

src/pages/generalSettings/GeneralSettings.jsx
Atticus29: Done.

PR #341, opened 2022-05-12
Dex 937 twitter handle user data

Atticus29:
schema.getValue → schema?.getValue in a few places, because it was breaking as soon as the user had a twitter handle but the twitterbot was disabled.

Added new data to the intelligentAgentSchema pertaining to user metadata.

Improved the setShowTwitterSuccess logic to be more robust to differences in the twitterStatusCode output vs the twitterTestResults output (and yes, the do disagree sometimes, resulting e.g. in situations where the status code is 400 but the test results still show success as true).

Added keys to IntelligentAgentSettings to make the warnings go away.

### DIFF
--- a/src/models/users/useUserMetadataSchemas.js
+++ b/src/models/users/useUserMetadataSchemas.js
@@ -50,6 +50,8 @@ export default function useUserMetadataSchemas(displayedUserId) {
             'data',
             'enablingField',
           ]);
+          console.log('deleteMe currentPlatformEnablingField is: ');
+          console.log(currentPlatformEnablingField);
           const isEnabled = get(siteSettings, [
             'data',
             currentPlatformEnablingField,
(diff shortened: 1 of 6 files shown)


src/models/users/useUserMetadataSchemas.js
```diff
@@ -31,6 +34,28 @@ export default function useUserMetadataSchemas(displayedUserId) {
             }),
           ]
         : [];
+      const intelligentAgentFields = map(
```
brmscheiner, line 37: is it just me or was this exact code in the last PR too?

brmscheiner (approved):

src/models/users/useUserMetadataSchemas.js
Atticus29: Very similar, but not exact.
Screen Shot 2022-05-12 at 10 22 20 AM

PR #345, opened 2022-05-14
Dex 937 placeholder

Atticus29:
Context: JH wasn't sure whether to put the @ or not in his twitter handle. So, now the label looks like this: 

Screen Shot 2022-05-13 at 7 15 33 PM

EditUserMetadata had to rely on the same label as Metadata card. In other words, the viewComponent and the editComponent for userMetadata were sharing a label.

Since TextFields don't seem to support a placeholder prop, I figured that changing the label when needed would suffice. Whether we want to call that placeholder or not is a good question.


Atticus29:
> Context: JH wasn't sure whether to put the @ or not in his twitter handle. So, now the label looks like this:
> 
> Screen Shot 2022-05-13 at 7 15 33 PM
> 
> EditUserMetadata had to rely on the same label as Metadata card. In other words, the viewComponent and the editComponent for userMetadata were sharing a label.
> 
> Since TextFields don't seem to support a placeholder prop, I figured that changing the label when needed would suffice. Whether we want to call that placeholder or not is a good question.

Another thought is trim()ming any initial @ character. We should probably do both.

src/components/fields/edit/TextEditor.jsx
```diff
@@ -37,7 +42,11 @@ const TextInput = function(props) {
         id={schema.name}
         multiline={isLongString}
         maxRows={isLongString ? 5 : undefined}
-        label={label}
+        label={
+          placeholder
```
brmscheiner, line 46: it doesn't really make sense to have a `placeholder` prop that takes precedence over label. if anything, i would expect placeholder to be a fallback if label was absent. probably the best way to do this is to change the labelId that is passed into the TextEditor component rather than changing TextEditor

brmscheiner (changes requested):

src/components/EditUserMetadata.jsx
```diff
@@ -123,12 +130,19 @@ export default function EditUserMetadata({
                 : omit(fieldValues, ['email']);
               const patchValues = map(
                 patchFieldValues,
-                (value, key) => ({
+                (value, key) => {
+                  if(key === twitterMetadataKey){
```
brmscheiner, line 134: looks like you need to run the formatter (`npm run format`)

src/components/fields/edit/TextEditor.jsx
```diff
@@ -17,7 +17,7 @@ const TextInput = function(props) {
     ...rest
   } = props;
 
-  const label = useLabel(schema);
+  const editLabelId = useEditLabel(schema);
```
brmscheiner, line 20: this returns a label not a labelId

src/hooks/useEditLabel.js
```diff
@@ -0,0 +1,17 @@
...
+import { get } from 'lodash-es';
+
+export default function useEditLabel(fieldSchema, noAsterisk = false) {
+  const intl = useIntl();
+  let label = null;
+  if (fieldSchema?.editLabelId) {
+    label = intl.formatMessage({ id: fieldSchema.editLabelId });
+  } else if(fieldSchema?.labelId){
+    label = intl.formatMessage({ id: fieldSchema.labelId });
+  }else
```
brmscheiner, line 11: dont forget to check for `editLabel`

brmscheiner: so it's a little weird to implement this in an abstract way (editLabel/viewLabel) but only to add functionality to the edit component TextEditor. the proper implementation of this should support all Edit and View components, otherwise it's just confusing

brmscheiner: like as a developer i would see the "viewLabelId" property is supported, then anywhere I tried to use it wouldn't work. does that make sense?

brmscheiner (changes requested):

src/components/EditUserMetadata.jsx
Atticus29: Done

src/components/fields/edit/TextEditor.jsx
Atticus29: Done

src/components/EditUserMetadata.jsx
```diff
@@ -22,9 +22,13 @@ function getInitialFormValues(schema) {
   }, {});
 }
 
-const twitterSchema = find(intelligentAgentSchema, schema => {
-  return schema?.platformName === 'twitter'
-},{});
+const twitterSchema = find(
```
brmscheiner, line 25: I'm not sure find accepts a fallback argument actually! So this might not behave as expected. From the docs:
```
_.find(collection, [predicate=_.identity], [fromIndex=0])
```

src/components/fields/edit/InputRow.jsx
```diff
@@ -1,7 +1,7 @@
 import React from 'react';
 import Skeleton from '@material-ui/lab/Skeleton';
 
-import useLabel from '../../../hooks/useLabel';
+import useViewLabel from '../../../hooks/useViewLabel';
```
brmscheiner, line 4: gah I know this is a maze, but `<InputRow>` is actually usually (or maybe even always) paired with edit components, not view components. So the behavior here is fine but actually the language is a little wonky.

src/hooks/useViewLabel.js
```diff
@@ -1,13 +1,17 @@
 import { useIntl } from 'react-intl';
 import { get } from 'lodash-es';
 
-export default function useLabel(fieldSchema, noAsterisk = false) {
+export default function useViewLabel(
```
brmscheiner, line 4: I think just renaming this to `useLabel` would do the trick!

brmscheiner (approved):

src/components/EditUserMetadata.jsx
Atticus29: Good catch!

PR #348, opened 2022-05-18
Disable kickoff annotation UI

brmscheiner:
Not removing the whole UI because I expect this functionality to be restored soon


Atticus29 (approved):
LGTM

PR #349, opened 2022-05-19
DEX-540 Add recaptcha to report sighting form

karenc:
Get the recaptcha public key from `/api/v1/site-settings/main/block`
(property `recaptchaPublicKey`).

Add
`<script id="recaptcha-script" src="https://www.google.com/recaptcha/api.js?render=${recaptchaPublicKey}"></script>`
to `<head>`.

Just before submitting the sighting, wait for `grecaptcha.ready`, then
get a token and send it with the sighting for back end verification.

### DIFF
--- a/src/pages/reportSighting/ReportForm.jsx
+++ b/src/pages/reportSighting/ReportForm.jsx
@@ -60,6 +60,19 @@ export default function ReportForm({
     data: siteSettingsData,
     siteSettingsVersion,
   } = useSiteSettings();
+
+  const recaptchaPublicKey = get(
+    siteSettingsData,
+    'recaptchaPublicKey',
+  ).value;
+  if (!document.getElementById('recaptcha-script')) {
+    const recaptchaApiUrl = `https://www.google.com/recaptcha/api.js?render=${recaptchaPublicKey}`;
+    const recaptchaScript = document.createElement('script');
+    recaptchaScript.src = recaptchaApiUrl;
+    recaptchaScript.id = 'recaptcha-script';
+    document.head.appendChild(recaptchaScript);
+  }
+
   const [sightingType, setSightingType] = useState(null);
 
   const {
@@ -313,6 +326,21 @@ export default function ReportForm({
                   ]),
                   sightings: [report],
                 };
+
+                const grecaptchaReady = () =>
+                  new Promise(resolve => {
+                    window.grecaptcha.ready(() => {
+                      resolve();
+                    });
+                  });
+
+                await grecaptchaReady();
+
+                const token = await window.grecaptcha.execute(
+                  recaptchaPublicKey,
+                  { action: 'submit' },
+                );
+                assetGroup.token = token;
 
                 const assetGroupData = await postAssetGroup(
                   assetGroup,


karenc:
By the way I totally expect my code to be in the wrong place or inefficient.  Please let me know how I should rewrite it :sweat_smile:

Atticus29 (dismissed):
LGTM

karenc:
@Atticus29 Could you also add a check for `recaptchaPublicKey` being null which means recaptcha isn't set up?  (so skip recaptcha)

Atticus29:
> @Atticus29 Could you also add a check for `recaptchaPublicKey` being null which means recaptcha isn't set up? (so skip recaptcha)

Ok @Emily-Ke @karenc , what do y'all think of this?

src/pages/reportSighting/ReportForm.jsx
```diff
@@ -314,6 +335,22 @@ export default function ReportForm({
                   sightings: [report],
                 };
 
+                const grecaptchaReady = new Promise(resolve => {
```
karenc, line 338: I think `if (recaptchaPublicKey)` needs to be above this line because if we don't include `recaptcha/api.js`, `window.grecaptcha` won't be defined

src/pages/reportSighting/ReportForm.jsx
```diff
@@ -314,6 +335,22 @@ export default function ReportForm({
...
+                  window.grecaptcha.ready(() => {
+                    resolve();
+                  });
+                });
+
+                await grecaptchaReady;
+
+                if (recaptchaPublicKey) {
```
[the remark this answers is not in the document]
karenc: Sounds good to me

Atticus29:
@karenc and @Emily-Ke I implemented the recommended change. 

I also noticed that Bulk Import was failing because of a recaptcha invalid input error (why? Users are logged in if they are able to access bulk import).

When I added the recaptcha code in to BulkReportForm, the error disappeared.

However, it ended up having to add the token property to the input object in the postAssetGroup call, which mean that that whole section of code had to be nested in if (window.grecaptcha) {.... This doesn't feel like a great idea, but I can't think of an alternative at the moment.

So, I guess my questions are:
1. Is there a way to make bulk import reporting not relianton recaptcha? And, if the answer is, "no",
2. Is my solution ok for the time being? If the answer is, "no", what you recommend I do instead?

karenc:
I just enabled recaptcha for `POST /api/v1/asset_groups/` without checking whether the user is checked in or not.  So for now we have to always send the token if recaptcha is configured in the back end.

I'm expecting the code to look something like this:

```javascript                   
            const assetGroup = {                   
              description: 'Bulk import from user',
              uploadType: 'bulk',                             
              speciesDetectionModel: [detectionModel || null],
              transactionId: get(assetReferences, [
                0,              
                'transactionId',
              ]),       
              sightings,
            };
                                    
            if (window.grecaptcha) {                          
              const grecaptchaReady = new Promise(resolve => {
                window.grecaptcha.ready(() => {
                  resolve();
                });
              });                   
              await grecaptchaReady;
                                                            
              const token = await window.grecaptcha.execute(
                recaptchaPublicKey,  
                { action: 'submit' },
              );                       
              assetGroup.token = token;
            }
                                                                    
            const assetGroupData = await postAssetGroup(assetGroup);
                                                             
            const assetGroupId = get(assetGroupData, 'guid');
            if (assetGroupId) {                                    
              history.push(`/bulk-import/success/${assetGroupId}`);
              queryClient.invalidateQueries(queryKeys.me);
   
```

karenc:
@Atticus29 ^ Notes about the last comment:

1. Pull the data out into a variable called `assetGroup`, this way we only need to add `assetGroup.token = token` inside the `if (window.grecaptcha)` block.
2. Put `grecaptchaReady` in the `if (window.grecaptcha)` block because it will error if `grecaptcha` isn't defined.

Emily-Ke:
Instead of repeating the same use effect, you can move it to a shared ancestor, ReportSightingsPage. Alternatively, you could extract it to a useRecaptcha hook.

Atticus29:
> ```js
> await postAssetGroup(
> ```

1. OMG of course. Please generously ignore the fact that I didn't think to do this.
2. Will do. Thanks!

Atticus29:
Alright. @karenc @Emily-Ke what do y'all think now. Sorry that this is taking more iterations than it ought 😊 .

src/pages/bulkImport/BulkReportForm.jsx
```diff
@@ -74,6 +75,10 @@ function onRecordInit(record) {
 export default function BulkReportForm({ assetReferences }) {
   const theme = useTheme();
   const history = useHistory();
+  const {
+    data: siteSettingsData,
+    siteSettingsVersion,
```
Emily-Ke, line 80: `siteSettingsVersion` is not used.

PR #355, opened 2022-05-23
Fix site settings version calculation

Emily-Ke:
The `siteSettingsVersion` was always undefined because it was being accessed from `data.response.configuration.data.response.version` instead of `data.response.version`.


Atticus29 (approved):
LGTM and definitely has a non-null version now.

brmscheiner (approved):

PR #356, opened 2022-05-23
Dex 1034 fix sex select and disambiguate duplicate custom field labels

Atticus29:
This PR solves a few issues:

1. Fixes some bugs with multi-select editor components that would crash when you tried to add more options to customFields that are multi-select in Control Panel → Manage Fields.

2. Removes src/models/assetGroup/useAssetGroupSchemas.js, which doesn't seem to get used anywhere.

3. Fixes a bug in the SaveField component where it should have been getting defaultValue from the schema instead of default from the formData object directly.

4. Fixes a bug where sex was a string field rather than a select field. Now, as a select field, bulk import entries of sex can be manually mapped to correct sex values during the bulk import process.

5. Fixes a bug where custom fields that were designated as select or multi-select were not being treated as such.

This PR does not solve the issue where multi-select custom fields can't be manually mapped to multiple values (see https://wildme.atlassian.net/browse/DEX-1080 for clarification about this issue. It has been separated out for the time being).
Screen Shot 2022-05-23 at 5 02 02 PM
Screen Shot 2022-05-23 at 5 03 19 PM

QA Notes:
After merge, you should now be able to: 
1. Designate sex manually during bulk import correctly (try maybe mis-spelling sex in  your bulk imports to test this).
2. Add options to multi-select and select customFields (e.g., Behavior at the encounter and at the sighting level).
3. See a distinction between custom fields that have the same label. For instance, zebra has a label, "Behavior" that is used for both the custom field at the encounter level and separately for a custom field at the sighting level. Previously, it would have been impossible to know which of the two "Behavior"s you were assigning to a given column during bulk import:
Screen Shot 2022-05-23 at 5 07 39 PM


src/components/fields/edit/SelectionEditor.jsx
```diff
@@ -23,9 +23,11 @@ const SelectionEditor = function(props) {
     ...rest
   } = props;
   const intl = useIntl();
+  const splitValues =
+    typeof value === 'string' ? value?.split(',') : value;
 
   function getLabel(object) {
-    if (object.labelId)
+    if (object?.labelId)
```
Atticus29, line 30: This gave me issues while I was testing, so I just fixed it.

src/components/fields/view/MultiSelectViewer.jsx
```diff
@@ -10,9 +10,11 @@ export default function MultiSelectViewer({
   defaultLabel = 'Option label not found',
 }) {
   const { choices } = schema;
+  const splitValues =
+    typeof value === 'string' ? value.split(',') : value;
 
-  const selectedOptions = choices.filter(choice =>
-    value.find(v => choice[lookupKey] === v),
+  const selectedOptions = choices?.filter(choice =>
```
Atticus29, line 16: Missing choices gave me issues while I was developing, so I just fixed it.

src/constants/sexOptions.js
```diff
@@ -2,18 +2,22 @@ export default [
   {
     value: 'male',
     labelId: 'MALE',
+    label: 'Male',
```
Atticus29, line 5: I wanted to just use labelId, but that seems to involve a refactor of createFieldSchema itself, which seemed out of scope of this already-creeping ticket.

src/pages/bulkImport/utils/useBulkImportFields.js
```diff
@@ -62,6 +66,13 @@ export default function useBulkImportFields() {
       );
       return bulkSightingFields.map(f => {
         const additionalProperties = {};
+        if (
+          f?.fieldType === 'select' ||
```
Atticus29, line 70: This will probably change a little bit when we get to DEX-1080.

src/pages/bulkImport/utils/useBulkImportFields.js
```diff
@@ -87,10 +98,24 @@ export default function useBulkImportFields() {
...
+          additionalProperties.type = 'select';
+        }
+        if (f?.choices) additionalProperties.options = f.choices;
         if (f.name === 'taxonomy') {
           additionalProperties.type = 'select';
           additionalProperties.options = speciesOptions;
         }
+        if (f.name === 'sex') {
+          const sanitizedSexOptions = Object.values(
+            omitBy(sexOptions, sex => sex?.labelId === 'BLANK'),
```
Atticus29, line 114: Had to remove Blank from the list, because a value of null was giving a type error.

src/pages/bulkImport/utils/useBulkImportFields.js
```diff
@@ -185,13 +210,42 @@ export default function useBulkImportFields() {
...
+    allAvailableFields,
+    field => field?.label,
+    [],
+  );
+  const duplicates = filter(
+    allAvailableFields,
+    (field, idx) => allLabels.indexOf(field?.label) !== idx,
+  );
+  let disambiguatedAvailableFields;
+  duplicates.forEach(duplicate => {
```
Atticus29, line 228: I figured that this forEach was ok, given that duplicates should be small and only a .map is called each time therein.

src/pages/bulkImport/utils/useBulkImportFields.js
```diff
@@ -62,6 +66,13 @@ export default function useBulkImportFields() {
       );
       return bulkSightingFields.map(f => {
         const additionalProperties = {};
+        if (
+          f?.fieldType === 'select' ||
+          f?.fieldType === 'multiselect'
+        ) {
+          additionalProperties.type = 'select';
```
Atticus29, line 73: select, checkbox, and string seem to be FlatFile-specific designations. I need to learn about whether they even accommodate multi-select in this way, and, if not, what we can do about it (treat it as a string and have some way to custom validate it?? Seems like a lot.)

src/utils/formatters.js
```diff
@@ -287,3 +287,20 @@ export const formatLocationFromSighting = (
 export const sanitizeTwitterHandle = value => {
   return value.replace(/^@/, '');
 };
+
+export const formatDuplicateLabel = (currentLabel, key, intl) => {
```
brmscheiner, line 291: this might never be used outside of bulk import. so instead of putting it in the common utils it might be better placed in the bulk import folder

brmscheiner (approved):

PR #357, opened 2022-05-24
Dex 665 show default custom values

Atticus29:
Default values for custom fields were not being displayed. Now, they are:
Screen Shot 2022-05-24 at 3 54 13 PM

### DIFF
--- a/src/components/cards/MetadataCard.jsx
+++ b/src/components/cards/MetadataCard.jsx
@@ -21,7 +21,12 @@ function fieldValueGood(field) {
   console.log('deleteMe and its value is: ');
   console.log(value);
   if (badValues.includes(value)) return false;
-  if (Array.isArray(value) && value?.length < 1) return false;
+  if (Array.isArray(value)) {
+    if (value?.length < 1) return false;
+    const removeNulls = value.filter(entry => entry !== null); //for lat/long
+    if (removeNulls?.length < 1) return false;
+  }
+
   if (value?.time === null && value?.timeSpecificity === null)
     return false;
   return true;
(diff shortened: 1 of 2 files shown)


src/components/cards/MetadataCard.jsx
```diff
@@ -17,6 +17,12 @@ const badValues = [null, undefined, ''];
 function fieldValueGood(field) {
   const value = field?.value;
   if (badValues.includes(value)) return false;
+  if (Array.isArray(value)) {
+    if (value?.length < 1) return false;
```
Atticus29, line 21: Multiselect custom fields come back with an empty array, which was returning true here.

src/components/cards/MetadataCard.jsx
```diff
@@ -17,6 +17,12 @@ const badValues = [null, undefined, ''];
...
+    const removeNulls = value.filter(entry => entry !== null);
+    if (removeNulls?.length < 1) return false;
```
Atticus29, line 23: This is for lat/long, although a legit 0,0 lat/long also comes back as [null, null] currently, so it'll be a blindspot for us.

brmscheiner (commented):
So this looks good, but I still think if a field is empty we should omit it from the metadata card. For example if default value is `''` we shouldn't display a title with nothing below it. What is the behavior in that situation?

brmscheiner (approved):

Atticus29:
> So this looks good, but I still think if a field is empty we should omit it from the metadata card. For example if default value is '' we shouldn't display a title with nothing below it. What is the behavior in that situation?

These won't be displayed because of the badValues logic:
```
const badValues = [null, undefined, ''];
function fieldValueGood(field) {
...
if (badValues.includes(value)) return false;
```

PR #359, opened 2022-05-26
ID pipeline status card design update

brmscheiner:
image

@Emily-Ke the purpose of this PR is to hand off designs for the pipeline status card to you. I added the image processing step, provided an action item for curation, and renamed matching to identification.

Feel free to merge this PR or not, it shouldn't be harmful but it's certainly not production ready yet either. My goal was to provide a clear path forward for how this should look, so let me know if there's any ambiguity remaining!


Emily-Ke:
I decided to use this as a template. Identification label changes are in #362

PR #360, opened 2022-05-26
Send re-run Id request

brmscheiner:
This PR adds the button to re-run identification on a sighting, but I am not seeing the any changes to the pipeline_status property after a successful POST to `/api/v1/sightings/8be7d16b-eb8f-4b6f-bec4-54d3daca0394/rerun_id`. Since Howard is out I wonder if someone else can look into it, maybe @naknomum or @bluemellophone?

### DIFF
--- a/src/pages/sighting/SightingEntityHeader.jsx
+++ b/src/pages/sighting/SightingEntityHeader.jsx
@@ -15,6 +15,7 @@ import FeaturedPhoto from '../../components/FeaturedPhoto';
 import { formatDateCustom } from '../../utils/formatters';
 import timePrecisionMap from '../../constants/timePrecisionMap';
 import defaultSightingSrc from '../../assets/defaultSighting.png';
+import RerunIdentificationDialog from './identification/RerunIdentificationDialog';
 
 const tabItems = [
   {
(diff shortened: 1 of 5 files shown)


src/pages/sighting/SightingEntityHeader.jsx
```diff
@@ -130,6 +137,12 @@ export default function SightingEntityHeader({
                   disabled: pending || sightingIsReviewed,
                   label: 'Mark sighting reviewed',
                 },
+                {
+                  id: 'rerun-identification',
+                  onClick: () => setRerunDialogOpen(true),
+                  disabled: pending,
```
brmscheiner, line 143: would like to add some more nuance here, like disable the button if identification is currently in progress.

Atticus29 (commented):
There are currently two interfaces that look like they ought to re-run identification:
Screen Shot 2022-05-27 at 1 48 07 PM
but also:
Screen Shot 2022-05-27 at 1 47 58 PM
The latter does nothing. Also, this kind of raises the bigger issue of, "why does re-run ID occur at the sighting level rather than at the individual level, given that a sighting can contain multiple individuals". Seems like a waste of resources to re-run ID on everyone (esp. if the sighting contains a whole pod/gaggle/murder, etc.).

I think that we should at least make the Actions menu re-run identification do something, or else get rid of it.

Atticus29:
Everything else made sense to me and looks good.

brmscheiner:
@Atticus29 nice catch that button has needed to be removed from the encounter for a while! i agree - running ID on just an encounter or an annotation is something we definitely want to support. when we do have the feature, we will try to encourage users to do that instead of re-running ID on the entire sighting.

Atticus29 (approved):
LGTM

PR #362, opened 2022-05-27
dex 967 - integrate new houston pipeline status api

Emily-Ke:
- Updates the sighting status card:
  -   Uses the new houston API. The intention is that any stage calculations will be done in houston rather than on the front end. This PR does not include any of the progress metrics. That will be handled in a future PR.
  - Changes the word "matching" to "identification"
  - Adds the image processing step
  - Updates the sighting submission message to handle plurals.

Also DEX-1081


src/pages/sighting/statusCard/StatusCard.jsx
```diff
@@ -39,58 +40,69 @@ function getProgressText(intl, startDate) {
   );
 }
 
+function getStage(pipelineStep) {
```
Atticus29, line 43: I think that this is beyond the scope of this ticket, but I see this getting lost if we ever need to report status in other places besides this card. I also think that there might be a way to generalize this a little more (if we end up adding more stages). But again, just pie-in-sky thoughts. Doesn't seem to be anything wrong with this current implementation.

src/pages/sighting/statusCard/StatusCard.jsx
```diff
@@ -104,13 +116,42 @@ export default function StatusCard({ sightingData }) {
...
+          stage={imageProcessingStage}
+          notStartedText={intl.formatMessage({
+            id: 'WAITING_ELLIPSES',
+          })}
+          inProgressText={getProgressText(
+            intl,
+            imageProcessingStartTime,
+            now,
+          )}
+          finishedText={`Image processing finished on ${getDateString(
```
Atticus29, line 131: En.jsonify?

Atticus29 (commented):
LGTM other than the times being off. If that's not in the purview of this ticket, I'll approve it.

src/pages/sighting/statusCard/StatusCard.jsx
Emily-Ke: Definitely. My plan was to leave as much as possible the way it was with the previous API response, so that the only change was how the stages were calculated. Obviously image processing wasn't a part of the previous response, but this matches the other timeline steps.

src/pages/sighting/statusCard/StatusCard.jsx
```diff
@@ -19,17 +20,17 @@ function getDateString(date) {
   return date ? formatDate(date, true) : 'unknown date';
 }
 
-function getProgressText(intl, startDate) {
+function getProgressText(intl, startDate, currentDate) {
```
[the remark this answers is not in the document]
Emily-Ke: I switched back from houston `now` to `new Date()` in https://github.com/WildMeOrg/codex-frontend/pull/362/commits/57c6cb8bce18e5456e076fca696a2849e6dbffdd. For me the Curation time elapsed is now correct, but Detection is incorrect because of the aforementioned missing time zone designator.

locale/en.json
```diff
@@ -131,10 +131,11 @@
   "NEW_INDIVIDUAL_CREATED_DESCRIPTION": "New individual created.",
   "PENDING_BULK_IMPORT_MESSAGE": "You reported {sightingCount} sightings on {date}. You must finish processing this bulk import before starting another.",
   "HISTORY": "History",
+  "IMAGE_PROCESSING_SKIPPED_MESSAGE": "Image processing skipped because no images were submitted.",
   "DETECTION_SKIPPED_MESSAGE": "Detection skipped, either there are no images or no detection model was specified.",
   "CURATION_SKIPPED_MESSAGE": "Curation skipped - no images to curate.",
-  "MATCHING_SKIPPED_MESSAGE": "Matching skipped - no images to match.",
-  "MATCHING": "Matching",
+  "IDENTIFICATION_SKIPPED_MESSAGE": "Identification skipped - no images to match.",
```
naknomum, line 137: there _might_ be other reasons id was skipped, so this message ("no images to match") might be confusing.
isnt there an option via the ui that the user can pick to skip id?

Emily-Ke: Yes. These messages need to be updated, but for this PR, I wanted to leave everything as it was except how the stages were calculated. This is just a change from "Matching" to "Identification" in the string.

naknomum (approved):
not seeing any big red flags, so approving -- with the caveat that i am not super clear on frontend world, of course.  :grin: 

one small note about wording on "skipped".

src/pages/sighting/statusCard/StatusCard.jsx
Emily-Ke: I'm going to leave this here, unchanged, for now because it sounds like it may still be in flux as the houston API becomes more solid.

Atticus29 (approved):
New changes LGTM! Well done!

locale/en.json
bluemellophone: I agree with Jon, but also recognize that this is something to improve on later.  I'm OK with this placeholder message for now.

PR #363, opened 2022-05-27
Cleanup placeholders - sighting history and site setup

Emily-Ke:
- Comments out the sighting history dialog to hide it from users while it is still using mock data.
- Replaces the dolphin image on the site setup page with Bao.

| before | after |
| ---       | ---    |
| before | after


brmscheiner (approved):
Thank you, don't forget to delete `newsite.png`!

brmscheiner:
Aha! Let's just merge this then.

PR #364, opened 2022-05-27
Refetch on mount

brmscheiner:


brmscheiner:
@Atticus29 this appears to resolve DEX-1097 so I assigned that ticket to myself and moved it to code review

Atticus29 (approved):
This makes sense to me (assuming that it doesn't have other unintended consequences), but FWIW, it doesn't seem to be resolving DEX-1097 (Creating an individual doesn't update that individual's name in the sightings associated with it)

PR #366, opened 2022-05-27
Basic image processing handling

Emily-Ke:
- Bulk import page
  - Adds polling for the asset group while image processing is in progress so that the UI can change when processing is complete.
  - Adds a simple image processing alert indicating that image processing is in progress and that some functionality is not available.
  - Sets the asset counts in the table to "pending".
- Pending sighting page
  - Adds the same basic image processing alert.
  - Hides all of the tabs while image processing is in progress.
  - Uses a Material UI Skeleton instead of a featured photo while image processing is in progress.
- Changes how the `activeTab` is calculated so that it is only ever a valid value.

These changes span both DEX-1083 and DEX-967

bulk-in-progress

-----

sighting-in-progress

*Please ignore the status card being incorrect on the sighting page. This will be fixed in #362*

### DIFF
--- a/src/pages/assetGroup/AssetGroup.jsx
+++ b/src/pages/assetGroup/AssetGroup.jsx
@@ -19,15 +19,37 @@ import Link from '../../components/Link';
 import MoreMenu from '../../components/MoreMenu';
 import ConfirmDelete from '../../components/ConfirmDelete';
 import EntityHeader from '../../components/EntityHeader';
+import CustomAlert from '../../components/Alert';
 import AGSTable from './AGSTable';
 
+const POLLING_INTERVAL = 5000; // 5 seconds
+
+function deriveRefetchInterval(resultData, query) {
+  const { complete, status } = {
+    ...resultData?.data?.progress_preparation,
+  };
+
+  const error = query.state?.error && { ...query.state.error };
+
+  const isProgressSettled =
+    complete || ['failed', 'cancelled'].includes(status) || error;
+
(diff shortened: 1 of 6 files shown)


src/pages/assetGroup/AssetGroup.jsx
```diff
@@ -19,15 +19,37 @@ import Link from '../../components/Link';
...
 import EntityHeader from '../../components/EntityHeader';
+import CustomAlert from '../../components/Alert';
 import AGSTable from './AGSTable';
 
+const POLLING_INTERVAL = 5000; // 5 seconds
+
+function deriveRefetchInterval(resultData, query) {
+  const { complete, status } = {
+    ...resultData?.data?.progress_preparation,
+  };
```
brmscheiner, line 30: why not just 
```
const { complete, status } = resultData?.data?.progress_preparation

or if that doesn't work 

const { complete, status } = resultData?.data?.progress_preparation || {}
```

brmscheiner (approved):
lgtm!

src/pages/assetGroup/AssetGroup.jsx
Emily-Ke: You're right. That's legacy from when I wasn't defining primitives.

Emily-Ke: Fixed in https://github.com/WildMeOrg/codex-frontend/pull/366/commits/49f3271d91e3f4cd39c39759146eaa482555bb08

PR #367, opened 2022-05-30
Dex 1066 recoverable user manager deleted collab

Atticus29:
# QA Notes

Mutually-revoked collaborations are no longer hidden from the collaboration table.

If either of the collaborations (user1 → user2 or user2 → user1) is revoked, this ticket makes the table display a restore icon instead of a delete icon.

Upon click, the collaboration is restored.

Screen Shot 2022-05-30 at 10 12 20 AM
Screen Shot 2022-05-30 at 10 12 28 AM

# Notes for code review

Note that useEstablishCollaborationAsUserManager does not work on collaborations that are in a revoked state already, so usePatchCollaboration was required instead. 

In fact, in order to restore the collaboration, two separate calls to usePatchCollaboration were required (one for each of the two collaborators), thereby necessitating: 
1. usage of Promise.all to handle the success logic and 
2. removal of onClearPatchSuccess and patchSuccess (because we are unable to distinguish between usePatchCollaboration calls for revocation and for restoration. 

Because of 2), I decided to handle the success logic with two useStates; it's a departure from our pattern of using the error and success handling built into usePost/useMutate, but I think that the departure is warranted given the uniqueness of needing usePatchCollaboration to do different things on this same component.


Emily-Ke (commented):
Currently if a restore fails, it will display an error message the the revocation failed.

src/components/ActionIcon.jsx
```diff
@@ -23,6 +24,10 @@ const variantMap = {
     labelId: 'MUTUAL_REVOKE',
     component: DeleteIcon,
   },
+  restore: {
+    labelId: 'RESTORE_COLLABORATION',
```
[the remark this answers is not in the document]
Atticus29: Done. I think "restore" is sufficient for the enduser to understand what's going on.

src/pages/userManagement/UserManagerCollaborationEditTable.jsx
```diff
@@ -182,30 +215,48 @@ export default function UserManagersCollaborationEditTable({
...
                 id: 'COLLAB_REVOKE_ERROR_SUPPLEMENTAL',
               })
             : intl.formatMessage({ id: 'UNKNOWN_ERROR' })}
         </CustomAlert>
       )}
-      {patchSuccess && (
+      {restoreError && (
+        <CustomAlert
+          severity="error"
+          titleId="COLLABORATION_REVOKE_ERROR"
```
Emily-Ke, line 236: This should be "COLLABORATION_RESTORE_ERROR", or something similar.

src/pages/userManagement/UserManagerCollaborationEditTable.jsx
```diff
@@ -182,30 +215,48 @@ export default function UserManagersCollaborationEditTable({
...
+          onClose={onClearRestoreError}
+        >
+          {restoreError
+            ? restoreError +
```
Emily-Ke, line 240: Same as above comment about `revokeError`.

src/pages/userManagement/UserManagerCollaborationEditTable.jsx
```diff
@@ -182,30 +215,48 @@ export default function UserManagersCollaborationEditTable({
...
           titleId="COLLABORATION_REVOKE_ERROR"
-          onClose={() => {
-            onClearPatchError();
-          }}
+          onClose={onClearRevokeError}
         >
-          {patchError
-            ? patchError +
+          {revokeError
+            ? revokeError +
```
Emily-Ke, line 225: I know that this logic wasn't introduced here, but, given the condition for this alert to render at all, isn't this redundant? I don't see how the unknown error message will ever be used.

Emily-Ke (changes requested):

src/pages/userManagement/UserManagerCollaborationEditTable.jsx
Atticus29: Resolved.

src/pages/userManagement/UserManagerCollaborationEditTable.jsx
Atticus29: Resolved. Good catch.

Emily-Ke (approved):
Looks good to me.

PR #370, opened 2022-05-30
Use site settings instead of apiKeys.js

karenc:
- Remove apiKeysTemplate.js
- Change code to use site settings instead of apiKeys
- The settings are `googleMapsApiKey`, `flatfileKey` and `sentryDsn`
- Use `fetch` to get the sentry dsn and move react render App into the `finally` handler

See WildMeOrg/houston#665

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ We are also looking for help from designers and 3D modelers! Please send an emai
 
 ## Development 
 
-To successfully run the project, you will need to recreate one file that is gitignored for security reasons, `/src/constants/apiKeys.js`. Follow the instructions in the template located at `/src/constants/apiKeysTemplate.js`. After that, just run
+Just run
 
 ```js
 npm install 
--- a/src/index.jsx
+++ b/src/index.jsx
@@ -2,6 +2,7 @@ import 'react-hot-loader/root'; // This needs to be imported before React.
 import React from 'react';
 import { render } from 'react-dom';
 import * as Sentry from '@sentry/react';
+import axios from 'axios';
 import { get } from 'lodash-es';
 
 import 'normalize.css';
@@ -11,10 +12,11 @@ import './styles/globalStyles.css';
 import pjson from '../package.json';
 import App from './App';
 
-fetch(`${__houston_url__}/api/v1/site-settings/main/block`)
-  .then(response => response.json())
-  .then(result => {
-    const sentryDsn = get(result, [
+axios
+  .get(`${__houston_url__}/api/v1/site-settings/main/sentryDsn`)
+  .then(response => {
+    const sentryDsn = get(response, [
+      'data',
       'response',
       'configuration',
       'sentryDsn',
--- a/src/index.jsx
+++ b/src/index.jsx
@@ -8,19 +8,29 @@ import 'normalize.css';
 import '../fonts/fonts.css';
 import './styles/globalStyles.css';
 
-import { sentryDsn } from './constants/apiKeys';
 import pjson from '../package.json';
 import App from './App';
 
-if (!__DEV__) {
-  Sentry.init({
-    dsn: sentryDsn,
-    release: `frontend@${get(pjson, 'release')}`,
+fetch(`${__houston_url__}/api/v1/site-settings/main/block`)
+  .then(response => response.json())
+  .then(result => {
+    const sentryDsn = get(result, [
+      'response',
+      'configuration',
+      'sentryDsn',
+      'value',
+    ]);
+    if (sentryDsn) {
+      Sentry.init({
+        dsn: sentryDsn,
+        release: `frontend@${get(pjson, 'release')}`,
+      });
+    }
+  })
+  .finally(() => {
+    const root = document.getElementById('root');
+
+    const load = () => render(<App />, root);
+
+    load();
   });
-}
-
-const root = document.getElementById('root');
-
-const load = () => render(<App />, root);
-
-load();
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -25,9 +25,6 @@ jobs:
       - name: Install dependencies
         run: npm install --legacy-peer-deps
 
-      - name: Copy template api keys file
-        run: cp src/constants/apiKeysTemplate.js src/constants/apiKeys.js
-
       - name: Compile build
         run: npm run build -- --env=houston=relative
 
--- a/.gitignore
+++ b/.gitignore
@@ -6,7 +6,6 @@ package-lock.json
 .DS_Store
 .vscode/
 
-/src/constants/apiKeys.js
 #: created on build
 /src/constants/version.js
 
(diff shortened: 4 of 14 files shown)


src/index.jsx
```diff
@@ -8,13 +8,12 @@ import 'normalize.css';
...
 import './styles/globalStyles.css';
 
-import { sentryDsn } from './constants/apiKeys';
 import pjson from '../package.json';
 import App from './App';
 
 if (!__DEV__) {
   Sentry.init({
-    dsn: sentryDsn,
+    dsn: 'FIXME-sentryDsn',
```
karenc, line 16: @Emily-Ke @Atticus29 I think it's not possible to get site settings here, what can I do instead?

config/webpack/webpack.common.js
```diff
@@ -35,6 +35,11 @@ module.exports = env => {
     __DEV__: isDev && !process.env.LINKED_DEV,
     __config__: JSON.stringify(config),
     __houston_url__: JSON.stringify(houstonUrl),
+    __sentry_dsn__: JSON.stringify(
+      isDev
+        ? process.env.SENTRY_DSN_DEVELOPMENT
+        : process.env.SENTRY_DSN_PRODUCTION,
```
Emily-Ke, line 41: Is there a houston PR that sets these environment variables?

karenc: There's no PR yet but the branch is [sentry-dsn](https://github.com/WildMeOrg/houston/tree/sentry-dsn) and this is the [commit](https://github.com/WildMeOrg/houston/commit/ab20bc9dae716ef27ca107d97271a3e45ce0f805):

```diff
commit ab20bc9dae716ef27ca107d97271a3e45ce0f805 (HEAD -> sentry-dsn, origin/sentry-dsn)
Author: karen chan <karen@karen-chan.com>
Date:   Wed Jun 1 13:48:55 2022 +0100

    Add sentry environment variables to dev-frontend docker-compose service
    
    We need to give dev-frontend the same environment variables as back end
    so both are using the same sentry dsn.

diff --git a/docker-compose.codex.yml b/docker-compose.codex.yml
index 26d7ab8e..798691a0 100644
--- a/docker-compose.codex.yml
+++ b/docker-compose.codex.yml
@@ -411,6 +411,8 @@ services:
       - intranet
     environment:
       HOUSTON_APP_CONTEXT: codex
+      SENTRY_DSN_DEVELOPMENT: "${SENTRY_DSN_DEVELOPMENT}"
+      SENTRY_DSN_PRODUCTION: "${SENTRY_DSN_PRODUCTION}"
       # See port served by 'www' component (i.e. the reverse proxy)
       HOST: "0.0.0.0"
       PORT: "84"
diff --git a/docker-compose.mws.yml b/docker-compose.mws.yml
index 99afb0c8..8820fb88 100644
--- a/docker-compose.mws.yml
+++ b/docker-compose.mws.yml
@@ -373,6 +373,8 @@ services:
       - intranet
     environment:
       HOUSTON_APP_CONTEXT: mws
+      SENTRY_DSN_DEVELOPMENT: "${SENTRY_DSN_DEVELOPMENT}"
+      SENTRY_DSN_PRODUCTION: "${SENTRY_DSN_PRODUCTION}"
       # See port served by 'www' component (i.e. the reverse proxy)
       HOST: "0.0.0.0"
       PORT: "84"
```

karenc:
@Emily-Ke I moved `Sentry.init` into `src/App.jsx` and actually tried sending errors to sentry.io and it's working.  I guess it won't catch the error if react fails to initialize but may be it's good enough?

src/index.jsx
```diff
@@ -8,19 +8,29 @@ import 'normalize.css';
...
 
-if (!__DEV__) {
-  Sentry.init({
-    dsn: sentryDsn,
-    release: `frontend@${get(pjson, 'release')}`,
-  });
-}
-
-const root = document.getElementById('root');
+fetch(`${__houston_url__}/api/v1/site-settings/main/block`)
```
karenc, line 14: @Emily-Ke Changed to use `fetch` and moved `load` into the `finally` handler.

Emily-Ke:
@karenc, I switched from `fetch` to `axios` because that it what we use for other requests. Also, I moved your logic to get the Google maps API key into a reusable function.

README.md
```diff
@@ -12,7 +12,7 @@ We are also looking for help from designers and 3D modelers! Please send an emai
 
 ## Development 
 
-To successfully run the project, you will need to recreate one file that is gitignored for security reasons, `/src/constants/apiKeys.js`. Follow the instructions in the template located at `/src/constants/apiKeysTemplate.js`. After that, just run
+Just run
```
Atticus29, line 15: Is the specification that api keys need to be in the .env file present in the houston docs? If not, it ought to be mentioned herein.

src/index.jsx
```diff
@@ -2,25 +2,39 @@ import 'react-hot-loader/root'; // This needs to be imported before React.
...
-if (!__DEV__) {
-  Sentry.init({
-    dsn: sentryDsn,
-    release: `frontend@${get(pjson, 'release')}`,
-  });
-}
+axios
+  .get(`${__houston_url__}/api/v1/site-settings/main/sentryDsn`)
+  .then(response => {
+    if (__DEV__) return;
```
Atticus29, line 18: Would we want to prevent the axios call if (__DEV__)?

src/index.jsx
```diff
@@ -2,25 +2,39 @@ import 'react-hot-loader/root'; // This needs to be imported before React.
...
 
-const root = document.getElementById('root');
+    const sentryDsn = get(response, [
```
Atticus29, line 20: I don't know enough about sentry to know how to determine whether this works as expected.

Atticus29 (commented):
Just a few minor comments.

README.md
karenc: I can't believe we don't have docs already about settings :cold_sweat: but here it is now https://github.com/WildMeOrg/houston/pull/702

src/index.jsx
karenc: We have two settings in houston `SENTRY_DSN_DEVELOPMENT` and `SENTRY_DSN_PRODUCTION` so I assume we want to use sentry whether it's development or not.  Let's remove `if (__DEV__) return;`

karenc:
@Emily-Ke Thanks for adding the `useGoogleMapsApiKey` hook and changing `fetch` to `axios`, the APIs are very similar.  I have removed `if (__DEV__) return;` and tested it and it's working with sentry :+1:

src/index.jsx
Emily-Ke: We were only initializing it in dev before:
https://github.com/WildMeOrg/codex-frontend/blob/a0a9a029c233a381f241a63d95031f6363e52c6b/src/index.jsx#L15-L20

karenc: Yes I know, but looking at the houston settings `SENTRY_DSN_DEVELOPMENT` and `SENTRY_DSN_PRODUCTION`, it's clear we want to send to sentry even when development mode is true. That's why I'm changing it. Houston is also sending errors to sentry when development mode is true.

README.md
Atticus29: Thank you!

PR #371, opened 2022-05-31
Dex 1005 fix multi select custom field edit

Atticus29:
This in conjunction with https://github.com/WildMeOrg/codex-frontend/pull/356 fixes the bug where users couldn't edit custom multi-select fields.

This current PR fixes the specific bug where the first option could not be created.

### DIFF
--- a/src/components/fields/edit/SelectionEditor.jsx
+++ b/src/components/fields/edit/SelectionEditor.jsx
@@ -22,18 +22,12 @@ const SelectionEditor = function(props) {
     minimalLabels = false,
     ...rest
   } = props;
-  console.log('deleteMe value is: ');
-  console.log(value);
   const intl = useIntl();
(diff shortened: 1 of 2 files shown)


src/components/fields/edit/SelectionEditor.jsx
```diff
@@ -23,19 +23,26 @@ const SelectionEditor = function(props) {
     ...rest
   } = props;
   const intl = useIntl();
-  const splitValues =
-    typeof value === 'string' ? value?.split(',') : value;
+  const isMultiselect = schema.fieldType === fieldTypes.multiselect;
+  const isAlreadyArray = Array.isArray(value);
+  const makeSafeArray = val => {
```
Atticus29, line 28: I imagine that there's a cleaner way to do this, but I couldn't immediately think of it.

Emily-Ke (approved):

PR #372, opened 2022-06-02
Replace placeholder notification image with letter

Emily-Ke:
This removes the placeholder image of Shane in the notifications panel. Because notifications don't currently provide an easy way to access the sender's profile image, I used the first letter of the sender's name in place of the image.

notification


src/components/AuthenticatedAppHeader/NotificationsPane.jsx
```diff
@@ -117,7 +116,7 @@ export default function NotificationsPane({
                     }}
                   >
                     <div style={{ display: 'flex' }}>
-                      <Avatar src={shane} variant="circular" />
+                      <Avatar>{user1Name[0].toUpperCase()}</Avatar>
```
Emily-Ke, line 119: `user1Name` should either be a non-empty value from the notification or "Unnamed User", so there should always be a first letter.
https://github.com/WildMeOrg/codex-frontend/blob/1fb6b1af5e84f92a6953666ff606826ed276e8c6/src/utils/notificationUtils.js#L10-L11

Atticus29 (approved):
I'll approve this, but user1Name doesn't always seem to map to the intended user.


For instance, if a user manager establishes a collaboration, I don't think there's a user1Name, so a U is displayed. Perhaps it's more accurate to display the user manager's initial as the avatar in this case?

Similarly, if user1 requests a collaboration, their avatar letter is displayed on user2's notification pane. However, when user2 approves, user1 sees their own initial in the approval notification.

Perhaps a more robust but burdensome solution involves designating in notificationSchema.js which, if any, notification prop from getNotificationProps should be used for the avatar?

But I'm happy making a separate ticket for this. What do you think?
Screen Shot 2022-06-03 at 5 24 57 PM

Emily-Ke:
We decided to accept these changes, and created DEX-1119 to handle showing the correct letter for each notification type.

Atticus29:
Posting the related JIRA ticket https://wildme.atlassian.net/browse/DEX-1119 here for posterity.

PR #373, opened 2022-06-02
Dex 929 follow up

Atticus29:
Ticket recommendations:

- Rename “Annotation” to “Candidate Rank”
- Rename “Match candidates” to “Candidate Annotations”
- Remove “C” from match candidate listings
- View Sighting column no longer sortable

Note that "Candidate Annotations" instead of Candidate annotations" seems to go against a style choice that Ben has  made app-wide, so I kept as, "Candidate annotations".

Also changed the Q1 to just 1 on the query table after a chat with Tanya about it.

Screen Shot 2022-06-02 at 4 11 35 PM

### DIFF
--- a/src/pages/match/MatchCandidatesTable.jsx
+++ b/src/pages/match/MatchCandidatesTable.jsx
@@ -7,9 +7,9 @@ import ActionIcon from '../../components/ActionIcon';
 const columns = [
   {
     name: 'index',
-    labelId: 'ANNOTATION',
+    labelId: 'CANDIDATE_RANK',
     options: {
-      customBodyRender: i => `C${i + 1}`,
+      customBodyRender: i => `${i + 1}`,
     },
   },
   {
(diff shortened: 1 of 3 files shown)


src/pages/match/MatchCandidatesTable.jsx
```diff
@@ -57,9 +58,8 @@ export default function MatchCandidatesTable({
   return (
     <DataDisplay
       idKey="guid"
-      title="Match candidates"
-      data={matchCandidates}
-      // variant="secondary"
+      titleId="CANDIDATE_ANNOTATIONS"
+      data={matchCandidates} // variant="secondary"
```
Emily-Ke, line 62: Did prettier move this comment?

Emily-Ke (approved):
Looks good to me. I agree with using sentence case instead of title case.

src/pages/match/MatchCandidatesTable.jsx
Atticus29: Yup.

PR #375, opened 2022-06-07
Dex 1108 metadata edit fix

Atticus29:
The problem was that metadata coming into the EditUserMetadata component included the object {value: undefined} when the twitterbot is not enabled.

In order to prevent a PATCH being send to path: /undefined, I simply filtered out objects from the metadata array that had one or fewer keys, instead of the much richer kosher entries from metadata.

### DIFF
--- a/src/components/EditUserMetadata.jsx
+++ b/src/components/EditUserMetadata.jsx
@@ -1,5 +1,5 @@
 import React, { useEffect, useState } from 'react';
-import { get, map, omit, find } from 'lodash-es';
+import { get, map, omit, find, filter } from 'lodash-es';
 
 import DialogContent from '@material-ui/core/DialogContent';
 import DialogActions from '@material-ui/core/DialogActions';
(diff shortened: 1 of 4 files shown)


src/components/EditUserMetadata.jsx
```diff
@@ -22,9 +22,10 @@ function getInitialFormValues(schema) {
   }, {});
 }
 
-const twitterSchema = find(intelligentAgentSchema, schema => {
-  return schema?.platformName === 'twitter';
-});
+const twitterSchema = find(
```
Atticus29, line 25: Must have been a prettier fix.

src/components/EditUserMetadata.jsx
```diff
@@ -46,7 +47,11 @@ export default function EditUserMetadata({
 
   useEffect(
     () => {
-      setFieldValues(getInitialFormValues(metadata));
+      const safeMetadata = filter(
+        metadata,
+        datum => Object.keys(datum).length > 1,
```
Atticus29, line 52: the malformed twitter stand-in metadata element had only one key called value. This seems like one way to remove it.

src/models/users/useUserMetadataSchemas.js
```diff
@@ -46,16 +46,18 @@ export default function useUserMetadataSchemas(displayedUserId) {
...
-        [],
+      );
+      const intelligentAgentFields = map(
+        enabledIntelligentAgentFields,
+        intelligentAgent =>
+          createFieldSchema(fieldTypes.string, {
+            name: intelligentAgent?.userMetadataKey,
+            labelId: intelligentAgent?.viewLabelId,
+            editLabelId: intelligentAgent?.editLabelId,
+            icon: intelligentAgent?.icon,
```
Emily-Ke, line 59: Is the optional chaining operator necessary?

Emily-Ke (approved):

src/models/users/useUserMetadataSchemas.js
Atticus29: I suppose not!

PR #274, opened 2022-03-09
Dex 771 remove encounter from individual

Atticus29:
Edit: THIS IS FOR DEX-770 AND DEX-771!

Handles deletion (nb: more than just detachment from the individual) of encounters from the "animals" tab of the sightings page.

Should delete and handle as normal except in the following cases:
1. If the encounter is the final one for a particular Individual, the user is prompted again confirm that they want to cascade the deletion to include deleting the individual.
2. If the encounter is the final one for the sighting, the user is prompted again confirm that they want to cascade the deletion to include deleting the sighting (and, potentially, individual). In this case, because the parent sighting is deleted, successful deletion should navigate the user back to the home page.

Handles "deletion" (this time simple decoupling of the encounter from the individual) on the individual's page in its "sightings" table upon click of the trashbin.

Should decouple as normal except in the following cases:
1. If it's the final encounter in the table (and therefore the final encounter associated with the individual), the button will be disabled.

Also worth noting that I don't know how to attach multiple encounters to the same individual currently given that individual search doesn't seem to be fully set up yet?

PR #376, opened 2022-06-08
Dex 1106 sighting cascade delete

Atticus29:
Deletion of a sighting could potentially orphan an individual. In order to prevent this from happening, we have a BE error that returns, in the case of a sighting deletion attempt, a vulnerable individual guid.

This PR leverages the existence of that guid and uses it to control the text of the end user's deletion prompt. It also allows for a user to proceed with the sighting deletion after acknowledging that it would also cause otherwise-orphaned individual(s) to also be deleted.

Note this this PR follows the pattern implemented in the encounter delete cascade (https://github.com/WildMeOrg/codex-frontend/pull/274).

QA Notes:
1. Deletion of a sighting with no individuals should proceed as expected.
2. Deletion of a sighting with only individual(s) who are linked to other sightings (i.e., to individuals who would not be orphaned) should proceed as expected.
3. Deletion attempt of a sighting that would orphan an individual(s) should not happen without a change in text in the confirmation dialog and an additional confirmation from the end user. If this second confirmation is made, the sighting should be deleted along with any otherwise orphaned individual(s). Note that I opted to keep the (now informative!) error in the confirmation dialog in this case:
Screen Shot 2022-06-08 at 3 44 01 PM
Screen Shot 2022-06-08 at 3 44 07 PM
Screen Shot 2022-06-08 at 3 44 11 PM


src/models/sighting/useDeleteSighting.js
```diff
@@ -7,14 +7,25 @@ export default function useDeleteSighting() {
...
   const [error, setError] = useState(null);
   const [success, setSuccess] = useState(false);
+  const [vulnerableIndividual, setVulnerableIndividual] = useState(
+    null,
+  );
 
-  const deleteSighting = async sightingId => {
+  const deleteSighting = async (
+    sightingId,
+    forceDeleteIndividual = false,
```
Emily-Ke, line 16: What do you think about using a config object here instead for the second argument?

src/models/sighting/useDeleteSighting.js
```diff
@@ -30,6 +41,10 @@ export default function useDeleteSighting() {
       return false;
     } catch (postError) {
       setLoading(false);
+      const vulnerableIndividualGuid =
+        postError?.response?.data?.vulnerableIndividualGuid;
+      if (vulnerableIndividualGuid)
+        setVulnerableIndividual(vulnerableIndividualGuid);
```
Emily-Ke, line 47: Is there a reason to set this conditionally rather than on every error. If the vulnerableIndividualGuid state has a value and a subsequent request returns an error without a `response?.data?.vulnerableIndividualGuid`, shouldn't the local state be cleared?

src/models/sighting/useDeleteSighting.js
```diff
@@ -44,5 +59,7 @@ export default function useDeleteSighting() {
     onClearError: () => setError(null),
     success,
     setSuccess,
+    vulnerableIndividual,
+    setVulnerableIndividual,
```
Emily-Ke, line 63: What about `clearVulnerableIndividual` instead? It would better match `onClearError` and for now all you are doing is clearing it anyway.

src/pages/sighting/SightingCore.jsx
```diff
@@ -79,6 +81,17 @@ export default function SightingCore({
...
+    messageForConfirmDelete,
+    setMessageForConfirmDelete,
+  ] = useState(null);
+
+  useEffect(() => {
+    const message = vulnerableIndividual
+      ? 'SIGHTING_DELETE_VULNERABLE_INDIVIDUAL_MESSAGE'
+      : 'CONFIRM_DELETE_SIGHTING_DESCRIPTION';
+    setMessageForConfirmDelete(message);
+  }, vulnerableIndividual);
```
Emily-Ke, line 94: The message does not need to be stored in state. It can just be calculated when the function runs.

src/pages/sighting/SightingCore.jsx
```diff
@@ -119,18 +132,29 @@ export default function SightingCore({
...
+            deleteResults = await deleteSighting(id, true);
+          }
+          if (!pending && !vulnerableIndividual) {
             deleteResults = await deleteSighting(id);
           }
           const successful = pending
             ? deleteResults?.status === 204
             : deleteResults;
           if (successful) {
+            setVulnerableIndividual(null);
```
Emily-Ke, line 157: I think it makes more sense to reset vulnerable individual in the `if(successful)` block of useDeleteSighting.
https://github.com/WildMeOrg/codex-frontend/blob/79b26e174c17de1512149c9f67ad257e88134bc5/src/models/sighting/useDeleteSighting.js#L32-L37

src/pages/sighting/SightingCore.jsx
```diff
@@ -146,16 +170,15 @@ export default function SightingCore({
...
       />
       <SightingEntityHeader
         activeTab={activeTab}
         data={data}
         loading={loading}
         pending={pending}
         preparing={isPreparationInProgress}
-        guid={id}
-        // setHistoryOpen={setHistoryOpen}
+        guid={id} // setHistoryOpen={setHistoryOpen}
```
Emily-Ke, line 181: For me:
1. prettier doesn't even run because there are existing lint violations in this file.
2. When I disable the lint violations, prettier still does not move this comment.

If possible, this comment should be left on the new line.

src/pages/sighting/SightingCore.jsx
```diff
@@ -119,18 +132,29 @@ export default function SightingCore({
       /> */}
       <ConfirmDelete
         open={deleteDialogOpen}
-        onClose={() => setDeleteDialogOpen(false)}
+        onClose={() => {
+          setMessageForConfirmDelete(
+            'CONFIRM_DELETE_SIGHTING_DESCRIPTION',
+          );
+          setVulnerableIndividual(null);
```
Emily-Ke, line 139: The message does not need to be set here. Clearing the vulnerable individual guid will cause the 'CONFIRM_DELETE_SIGHTING_DESCRIPTION' message to be used.

src/pages/sighting/SightingCore.jsx
```diff
@@ -119,18 +132,29 @@ export default function SightingCore({
       /> */}
       <ConfirmDelete
         open={deleteDialogOpen}
-        onClose={() => setDeleteDialogOpen(false)}
+        onClose={() => {
```
Emily-Ke, line 135: The error should probably be cleared here too. Currently if the dialog is closed and re-opened, the message reverts to the sighting deletion confirmation, but the error alert is still visible.

Emily-Ke (changes requested):

src/models/sighting/useDeleteSighting.js
Atticus29: Could you elaborate?

src/models/sighting/useDeleteSighting.js
Atticus29: You're totally right.

src/models/sighting/useDeleteSighting.js
Atticus29: Good point!

src/pages/sighting/SightingCore.jsx
Atticus29: Nice catch.

Atticus29: Good point.

src/pages/sighting/SightingCore.jsx
Atticus29: Good point.

src/pages/sighting/SightingCore.jsx
Atticus29: ... I have no idea what happened here, but husky/prettier/etc. all seem fine _either way_ with me. Hmm. Setting to its own line...

src/pages/sighting/SightingCore.jsx
Atticus29: Ok. Done.

src/pages/sighting/SightingCore.jsx
Atticus29: I was still having trouble wrapping my head around react lifecycle stuff sometimes. It clearly works without storing in state, exactly as you've indicated here. Thank you for prompting me to re-read a little bit about when components re-render lol. I feel more comfortable with why this was unnecessary now.

src/models/sighting/useDeleteSighting.js
Emily-Ke: something like
```js
const deleteSighting = async (sightingId, config = {}) {
  const { forceDeleteIndividual = false } = config;
  //...
}
```
which is similar to https://github.com/WildMeOrg/codex-frontend/blob/d8a689389916b96156f9b4858a2311cabffc4716/src/models/sighting/useSighting.js#L4

It is more verbose than what you have though.

src/pages/sighting/SightingCore.jsx
```diff
@@ -119,12 +125,20 @@ export default function SightingCore({
...
         onDelete={async () => {
           let deleteResults;
           if (pending) {
             deleteResults = await deleteAssetGroupSighting(id);
-          } else {
+          }
+          if (!pending && vulnerableIndividual) {
+            deleteResults = await deleteSighting(id, true);
+          }
+          if (!pending && !vulnerableIndividual) {
```
Emily-Ke, line 141:
```js
if (pending) {
  // ...
} else if (vulnerableIndividual) {
  // ...
} else {
  // ...
}
```

src/pages/sighting/SightingCore.jsx
```diff
@@ -64,6 +66,10 @@ export default function SightingCore({
     onClearError: deleteAsgOnClearError,
   } = useDeleteAssetGroupSighting();
 
+  const onClearError = pending
+    ? deleteAsgOnClearError
+    : deleteSightingOnClearError;
```
Emily-Ke, line 71: This should also be used here
https://github.com/WildMeOrg/codex-frontend/blob/d8a689389916b96156f9b4858a2311cabffc4716/src/pages/sighting/SightingCore.jsx#L146-L148

src/components/ConfirmDelete.jsx
```diff
@@ -22,6 +22,7 @@ export default function ConfirmDelete({
   errorTitleId = 'SERVER_ERROR',
   error,
   titleId,
+  severity = 'error',
```
Emily-Ke, line 25: What do you think of `alertSeverity` instead?

src/pages/sighting/SightingCore.jsx
```diff
@@ -143,10 +157,20 @@ export default function SightingCore({
             ? deleteAssetGroupSightingError
             : deleteSightingError
         }
+        errorTitleId={
+          vulnerableIndividual
+            ? 'REQUEST_REQUIRES_ADDITIONAL_CONFIRMATION'
+            : 'SERVER_ERROR'
+        }
```
Emily-Ke, line 164: You can get the default `errorTitleId` when there is no vulnerable individual and still have the custom `errorTitleId` when there is by changing this to
```js
errorTitleId={
  vulnerableIndividual
    ? 'REQUEST_REQUIRES_ADDITIONAL_CONFIRMATION'
    : undefined
}
```

Emily-Ke (approved):

src/pages/sighting/SightingCore.jsx
Atticus29: Good catch. Done.

src/pages/sighting/SightingCore.jsx
Atticus29: Good catch. Done.

src/components/ConfirmDelete.jsx
Atticus29: Done.

src/pages/sighting/SightingCore.jsx
Atticus29: I'm surprised by this. Thank you. Done.

Atticus29:
Screen Shot 2022-06-11 at 3 52 32 PM
I ended up trying this out (I recorded my progress in the branch DEX-1106-follow-up), but hit some snags when onClearError now needed to deal with clearing an error statement that didn't come straight from the custom hook. Totally solvable, but I'll leave it for another day when I can address both it and bringing encounter delete into the same pattern.

Atticus29:
Went ahead and made https://wildme.atlassian.net/browse/DEX-1168 for that follow-up work.

PR #377, opened 2022-06-09
Dex 1091 quieter twitterbot

Atticus29:
This PR fixes the issue where useGetTwitterbotTestResults was being called despite having disabled passed to it as a queryOption.

I tried a number of things such as using enabled instead of disabled as the query option. These did not solve the issue (at least, they were not sufficient to solve the issue). However, I ended up using fewer double negatives in the codebase, so I decided to keep using enabled.

What I believe was the issue was gleaned from [this page](https://github.com/TanStack/query/issues/1196):

> Maybe you were setting enabled to a falsy value rather than really setting it to false ? This is a breaking change in v3 (stated here: https://react-query-beta.tanstack.com/guides/migrating-to-react-query-3#queryoptionsenabled). Falsy values will no longer work, it needs to really be the boolean value false.

So, despite `get(currentValues, 'intelligent_agent_twitterbot_enabled')` seeming to fetch a boolean value:
Screen Shot 2022-06-08 at 6 49 47 PM

, the issue disappeared as soon as I typecast to a Boolean explicitly with:

```
const isTwitterEnabled = Boolean(
    get(currentValues, 'intelligent_agent_twitterbot_enabled'),
  );
```
I do not understand why this seems to have been necessary.

I also noticed that `queryOptions.disabled` is being used to handle some `displayedLoading` logic in useFetch. I could not find any documentation about disabled as a queryOption; could this be a bug? If not, I'll remove the comment before merging.

To QA:
Open the network tab while loading /admin/settings. You should no longer see calls to intelligent_agent_twitterbot like this:

Screen Shot 2022-06-08 at 6 52 02 PM

If you still see it, make sure that you have cleared out the network tab's calls before reloading the page.

### DIFF
--- a/src/hooks/useFetch.js
+++ b/src/hooks/useFetch.js
@@ -30,7 +30,7 @@ export default function useFetch({
 }) {
   const [displayedError, setDisplayedError] = useState(null);
   const [displayedLoading, setDisplayedLoading] = useState(
-    !queryOptions.disabled,
+    !queryOptions.disabled, // should this use enabled instead of disabled? I couldn't find anything in the react-query documentation about disabled.
   );
   const [statusCode, setStatusCode] = useState(null);
 
--- a/src/models/site/useGetTwitterbotTestResults.js
+++ b/src/models/site/useGetTwitterbotTestResults.js
@@ -1,13 +1,13 @@
 import queryKeys from '../../constants/queryKeys';
 import useFetch from '../../hooks/useFetch';
 
-export default function useGetTwitterbotTestResults(disabled) {
+export default function useGetTwitterbotTestResults(enabled) {
   return useFetch({
     queryKey: queryKeys.twitterBotTestResults,
     url: '/site-settings/test/intelligent_agent_twitterbot',
     queryOptions: {
       retry: 1,
-      disabled,
+      enabled,
     },
   });
 }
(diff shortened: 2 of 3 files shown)


src/hooks/useFetch.js
```diff
@@ -30,7 +30,7 @@ export default function useFetch({
 }) {
   const [displayedError, setDisplayedError] = useState(null);
   const [displayedLoading, setDisplayedLoading] = useState(
-    !queryOptions.disabled,
+    !queryOptions.disabled, // should this use enabled instead of disabled? I couldn't find anything in the react-query documentation about disabled.
```
Emily-Ke, line 33: It looks like it was added in #330 when `config.disabled` was added to some of the useFetch derived hooks. Even those hooks set `queryOptions.enabled` based on that config, so it doesn't look like `queryOptions.disabled` is ever set to true. Please correct me on this if you find a case where it is.

That means that this line doesn't need to be `queryOptions.enabled`, but can just default to `true` since that is how it always currently evaluates anyway.

Emily-Ke (approved):
Looks good to me.

Emily-Ke:
`get(currentValues, 'intelligent_agent_twitterbot_enabled')` initially returns undefined because currentValues defaults to null.
https://github.com/WildMeOrg/codex-frontend/blob/2557280c88e7cb23f83ac4a29cc5aed8f6011f97/src/pages/generalSettings/GeneralSettings.jsx#L69
`get(currentValues, 'intelligent_agent_twitterbot_enabled', false)`, which utilizes the default value for `get` would most likely have solved the problem too.

src/hooks/useFetch.js
Atticus29: I also didn't find anything anywhere else.
I'm still a bit confused about what this does in practical terms, but I'll pull my comment out in the meantime.

PR #378, opened 2022-06-10
dex 1083 - improved preparation state feedback for bulk import 

Emily-Ke:
- Adds an error alert message.
- Adds an indeterminate progress bar to the in progress alert so that the page gives a slightly better impression that it is doing something.
  - eventually there will be metrics included in the alert and the progress bar will probably move into the alert too. This is a quick, intermediary change that hopefully improves the user experience slightly.

error-message

progress-bar

## QA notes
- A large bulk import will show the in progress state naturally. You can force that state by stopping all of your celery containers.
- To force the error state:
  - stop your celery containers
  - add something that will error to the first line of the set function
https://github.com/WildMeOrg/houston/blob/a3dba0dd4804c1439aae90bc1ca08cc3896e5eac/app/modules/progress/models.py#L419
  - start your celery containers


Atticus29 (approved):
LGTM. I didn't reproduce the errors because you outlined that in the QA notes and the code was quite clear.

PR #380, opened 2022-06-10
dex 1148 - update matching set query to use macros

Emily-Ke:
The front end implementation of WildMeOrg/houston#704.

This also changes the locationID query from a `term` to a `match_phrase` to enable the GUID search.


naknomum (approved):
these look good to me and are in sync with backend DEX-1149

brmscheiner (approved):

PR #381, opened 2022-06-13
Exclude individual from forming a relationship with itself

brmscheiner:
Casual encounters are still permitted


Atticus29 (approved):
Nice and easy fix. Works!

PR #382, opened 2022-06-13
dex 967 - update AGS and sighting status messages

Emily-Ke:
- Removes the photo count from the "Sighted Reported" stage because that information is not available until "Sighting Preparation" has completed.
- For the "Sighted Reported" date, uses the new `submissionTime` instead of `createdHouston` property. WildmeOrg/houston#678
- Replaces "Image Processing" with "Sighting Preparation" because every sighting goes through the preparation step. Even when there are no images reported, metadata files are created and other tasks are handled.
  - For this reason, this step should never have a value of `skipped: true`. A basic message for this case is still provided as a fallback.

| original | updated |
| :---: | :---: |
| image-processing | sighting-preparation |
- Splits "Skipped" messages into a default, and messages with more information about why the step was skipped.
- Internationalizes the "Finished" messages.


brmscheiner (approved):
lgtm!

PR #383, opened 2022-06-13
Dex 1119 notification placeholder improvement

Atticus29:
Adds a notificationAvatar to each notification schema to indicate which variable from NotificationUtils's getNotificationProps to use for gettin the first letter as an Avatar stand-in. 

Also fixes a bug that we missed earlier when collaboration_revoked was changed to collaboration_denied.

QA notes:
Here's what it looked like before (Note all of the "U"s as the avatar):
Screen Shot 2022-06-13 at 2 17 03 PM
Screen Shot 2022-06-13 at 2 17 51 PM
Screen Shot 2022-06-13 at 2 17 56 PM

Here's what is looks like after the fix:
Screen Shot 2022-06-13 at 2 21 02 PM
Screen Shot 2022-06-13 at 2 22 52 PM:
Screen Shot 2022-06-13 at 2 20 19 PM

Note that there's about an hour's worth of state built up on unacceptablebehavior.org right now if you log in as user1@example.com with the default password, so you don't have to reproduce all of these notifications yourself.

### Update Jun 14, 2022:

collaboration_denied and
collaboration_revoke

are both accommodated now.
See Jerome Jerome The Metronome and Betty pictured below:
Screen Shot 2022-06-14 at 11 37 41 AM
Screen Shot 2022-06-14 at 11 37 47 AM

### DIFF
--- a/src/components/AuthenticatedAppHeader/NotificationsPane.jsx
+++ b/src/components/AuthenticatedAppHeader/NotificationsPane.jsx
@@ -106,6 +106,10 @@ export default function NotificationsPane({
               const timeSince = calculatePrettyTimeElapsedSince(
                 createdDate,
               );
+              const avatarName =
+                eval(notification?.notificationAvatar) || user1Name;
+              console.log('deleteMe avatarName is: ');
+              console.log(avatarName);
               return (
                 <React.Fragment key={notification.guid}>
                   <Grid
@@ -118,7 +122,7 @@ export default function NotificationsPane({
                     }}
                   >
                     <div style={{ display: 'flex' }}>
-                      <Avatar>{user1Name[0].toUpperCase()}</Avatar>
+                      <Avatar>{avatarName[0].toUpperCase()}</Avatar>
                       <NotificationPaneDisplayText
                         currentNotificationSchema={
                           currentNotificationSchema
@@ -190,7 +194,9 @@ export default function NotificationsPane({
                     }}
                   >
                     <Text
-                      style={{ color: theme.palette.text.secondary }}
+                      style={{
+                        color: theme.palette.text.secondary,
+                      }}
                     >
                       {intl.formatMessage(
                         { id: 'TIME_SINCE' },
--- a/src/components/AuthenticatedAppHeader/NotificationsPane.jsx
+++ b/src/components/AuthenticatedAppHeader/NotificationsPane.jsx
@@ -107,7 +107,9 @@ export default function NotificationsPane({
                 createdDate,
               );
               const avatarName =
-                `${notification?.notificationAvatar}` || user1Name;
+                getNotificationProps[
+                  (notification?.notificationAvatar)
+                ] || user1Name;
               console.log('deleteMe avatarName is: ');
               console.log(avatarName);
               return (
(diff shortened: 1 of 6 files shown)


brmscheiner (approved):

src/components/AuthenticatedAppHeader/NotificationsPane.jsx
```diff
@@ -104,6 +104,11 @@ export default function NotificationsPane({
               const timeSince = calculatePrettyTimeElapsedSince(
                 createdDate,
               );
+
+              const avatarName =
+                getNotificationProps(notification)[
```
Emily-Ke, line 109: `getNotificationProps` has already been run.
This can be changed to:
```
const notificationProps = getNotificationProps(notification)
const {userName, userNameGuid, // everything else being destructured } = notificaitonProps;
//...
const avatarName = notificationProps[currentNotificationSchema?.notificationAvatar] || 'Unnamed User'  // no extra parenthesis needed either
```

src/constants/notificationSchema.js
```diff
@@ -22,6 +22,7 @@ notificationSchemaPlaceholder[
     'A_COLLABORATION_WAS_CREATED_ON_YOUR_BEHALF_MORE_DETAILED',
   showNotificationDialog: false,
   buttonPath: '/#collab-card',
+  notificationAvatar: 'userName',
```
Emily-Ke, line 25: Maybe something like `notificationAvatarKey` or `notificationAvatarField` is clearer?

Emily-Ke (commented):
All of the `notificationAvatar`'s are 'userName'. Is that correct? If so, `notificaitonAvatar` is not necessary.

src/constants/notificationSchema.js
```diff
@@ -2,7 +2,7 @@ export const notificationTypeNames = {
   collaboration_manager_create: 'collaboration_manager_create',
   collaboration_request: 'collaboration_request',
   collaboration_approved: 'collaboration_approved',
-  collaboration_revoke: 'collaboration_revoke',
+  collaboration_revoke: 'collaboration_denied',
```
Emily-Ke, line 5: Are you sure about this? Both of them are houston Notification types
https://github.com/WildMeOrg/houston/blob/923dbab77fc1bb1c6105ec7d2b32bb7860c9f78b/app/modules/notifications/models.py#L18

src/components/AuthenticatedAppHeader/NotificationsPane.jsx
Atticus29: I didn't know that I could do this. Makes complete sense now. Worldview re-calibrated lol.

Atticus29: Good call.

src/constants/notificationSchema.js
Atticus29: Done.

src/constants/notificationSchema.js
Atticus29: Good catch. Both should exist. Should be implemented now.

Emily-Ke (approved):
Looks good to me too.

PR #385, opened 2022-06-14
Name changes and a bugfix 

brmscheiner:
splash page -> front page 
forum username -> community username 

also noticed that there were errors when dismissing the success alert on the splash settings page. so that is fixed here too


Atticus29 (approved):
LGTM

PR #386, opened 2022-06-14
dex 1012 - Add migration specific status messages

Emily-Ke:
- Adds migration information to the "Curation" finished message and "Identification" skipped message
migration-status


locale/en.json
```diff
@@ -142,6 +142,7 @@
   "CURATION_SKIPPED_MESSAGE": "Curation skipped - no images to curate.",
   "IDENTIFICATION_SKIPPED_MESSAGE": "Identification skipped.",
   "IDENTIFICATION_SKIPPED_NO_IMAGES_MESSAGE": "Identification skipped - no images to match.",
+  "IDENTIFICATION_SKIPPED_MIGRATED_MESSAGE": "Identification skipped - migrated sighting.",
```
brmscheiner, line 145: How about "Identification skipped - this sighting was migrated from the Zebra Wildbook" ?

brmscheiner (approved):

locale/en.json
Emily-Ke: Is there an instance ID that we have access to that we could use to know that this instance is zebra?

brmscheiner: Ahh, no. How about "Identification skipped - this sighting was migrated from a Wildbook database."

src/constants/wildbookSystemGuids.js
```diff
@@ -0,0 +1,5 @@
+const wildbookSystemGuids = {
+  'e144fec3-2f47-4195-8025-a8915f9a52a1': 'Wildbook for Zebras',
```
Emily-Ke, line 2: This is the system GUID for zebra.wildme.org. It is  a site-setting that is a unique guid per instance.

Emily-Ke:
There is a GUID that identifies a unique Codex instance, so on production instances, we should be able to have more specific migrated messages:

**Production migration (hopefully)**
migration-specific

**Non-production migration**
migration-general

**Non-migration**
default

src/constants/wildbookSystemGuids.js
brmscheiner: would this be a `wildbookSystemGuid` or a `codexSystemGuid`?

brmscheiner (approved):

src/constants/wildbookSystemGuids.js
Emily-Ke: It is a codex system GUID. What do you think of changing the name to `wildbooksBySystemGuids` then?

If we thought we were going to store other information too, I could change it too something like
```
const systemInformation = {
  'e144fec3-2f47-4195-8025-a8915f9a52a1': {
    migratedFrom: 'Wildbook for Zebras',
    // ... other information
  }
};
```

Perhaps this isn't even data that the front end should be storing, but something that we should request from houston.

Emily-Ke: Renamed to `wildbookBySystemGuid` in https://github.com/WildMeOrg/codex-frontend/pull/386/commits/f346ab342487247e98ac4b0799299d54cd94bbaa

PR #388, opened 2022-06-16
dex-1178 Update landing page metrics to current zebra values

Emily-Ke:
- Updates the hard coded metrics to the current values on zebra.wildme.org:
  -  Identified Individuals - the number of individuals
  - Reported Sightings - the number of asset group sightings plus the number of migrated sightings (sightings without asset group sighting GUIDs)
  - Users - the number of users that are not `internal` an do not have an `@localhost` email address
- Adds `FormattedNumber` so that the counts are formatted.

**NOTE**: This is a temporary fix until these values can be accessed via houston.
metrics

### DIFF
--- a/src/pages/splash/Metrics.jsx
+++ b/src/pages/splash/Metrics.jsx
@@ -1,4 +1,5 @@
 import React from 'react';
+import { FormattedNumber } from 'react-intl';
 import { useTheme } from '@material-ui/core/styles';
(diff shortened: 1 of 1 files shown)


Atticus29 (approved):
LGTM. I would have excluded @wildme.Org and perhaps @example.com. But that's so small, it's probably best to fry bigger fish.

PR #390, opened 2022-06-17
Dex 1196 id results before available fix

Atticus29:
When match results are not yet ready, clicking "view identification results" from the "Animal" tab's encounter(s) takes the user to the overview tab, where they can see that ID is still in progress/failed.

When match results are ready, the same button takes the user directly to the match results page.

JH mentioned casually in my 1:1 today that ideally he'd want the click to navigate directly to focus on the identification step.
I tried a few things to this effect (e.g., https://github.com/WildMeOrg/codex-frontend/pull/390/commits/d4fe72ca7f900d3d1b7c90b357b86d0b35345e21 and https://github.com/WildMeOrg/codex-frontend/pull/390/commits/1821b907f73cc185099602a2755a55dd90965880), and none of them seemed to work. Open to suggestions.

I'm gonna leave the domId in there even though I didn't end up using it; it's not hurting anything, and we may eventually want it.

Update: June 17 - The ticket was slightly redefined, and I removed domId in accordance with best practice.

### DIFF
--- a/src/pages/sighting/encounters/Encounters.jsx
+++ b/src/pages/sighting/encounters/Encounters.jsx
@@ -98,6 +98,16 @@ export default function Encounters({
   const encounterFieldSchemas = useEncounterFieldSchemas();
   const encounters = get(sightingData, 'encounters', []);
 
+  const { identification: identificationStep } =
+    sightingData?.pipeline_status || {};
+
+  const {
+    complete: isIdentificationComplete,
+    failed: isIdentificationFailed,
+  } = identificationStep || {};
+  const isIdReady =
+    isIdentificationComplete && !isIdentificationFailed;
+
   useEffect(
     () => {
       const message = vulnerableObject
@@ -222,7 +232,9 @@ export default function Encounters({
           {
             id: 'view-id-results',
             labelId: 'VIEW_IDENTIFICATION_RESULTS',
-            href: `/match-results/${sightingId}`,
+            href: isIdReady
+              ? `/match-results/${sightingId}`
+              : '#overview',
           },
           {
             id: 'create-new-individual',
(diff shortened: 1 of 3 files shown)


src/pages/sighting/encounters/Encounters.jsx
```diff
@@ -219,11 +229,13 @@ export default function Encounters({
...
-            labelId: 'VIEW_IDENTIFICATION_RESULTS',
-            href: `/match-results/${sightingId}`,
-          },
+          isIdReady
+            ? {
+                id: 'view-id-results',
+                labelId: 'VIEW_IDENTIFICATION_RESULTS',
+                href: `/match-results/${sightingId}`,
+              }
+            : {},
```
brmscheiner, line 238: we shouldn't pass an empty object here, just omit the object if ~isIdReady

brmscheiner (changes requested):
There is another view identification results button in the sighting status card

Atticus29:
>  There is another view identification results button in the sighting status card

This was completed in a previous PR:
Screen Shot 2022-06-22 at 10 51 48 AM
Screen Shot 2022-06-22 at 10 52 21 AM

src/pages/sighting/encounters/Encounters.jsx
Atticus29: @brmscheiner what do you think of this solution: https://github.com/WildMeOrg/codex-frontend/pull/390/commits/38faa05999e3185f0946071c8f506b860292b130?

I was also considering just splicing the object into an already-existing array if isIdReady, but that seemed less elegant.

Did you have another solution in mind?

brmscheiner (approved):

Emily-Ke (approved):

PR #391, opened 2022-06-17
Table cleanup!

brmscheiner:


src/constants/fieldCategories.js
```diff
@@ -26,7 +26,7 @@ export const defaultSightingCategories = {
 
 export const defaultEncounterCategories = {
   animal: {
-    name: 'individual',
+    name: 'individualInformation',
```
brmscheiner, line 29: mediocre names were causing console warnings for duplicate keys

src/pages/userManagement/UserEditTable.jsx
```diff
@@ -116,10 +115,8 @@ export default function UserEditTable() {
         variant="secondary"
         columns={tableColumns}
         data={activeUsers}
+        loading={loading}
```
brmscheiner, line 118: resolves dex-978

src/pages/userManagement/UserManagerCollaborationEditTable.jsx
```diff
@@ -39,14 +34,10 @@ export default function UserManagersCollaborationEditTable({
     mutate: restoreCollab,
     success: restoreSuccess,
     clearSuccess: clearRestoreSuccess,
-    loading: restoreLoading,
     error: restoreError,
     clearError: onClearRestoreError,
   } = usePatchCollaboration();
 
-  const isLoading =
-    userDataLoading || revokeLoading || restoreLoading;
```
brmscheiner, line 48: currently, this table doesn't use a loading state or update the data after revoke or restore actions are clicked. eventually these loading statuses will need to be used but right now they were just causing a grey box to appear below the table.

src/constants/fieldCategories.js
Emily-Ke: It doesn't look like 'individual' was causing duplicate keys. I agree with updating it for consistency with the other categories though.

src/pages/fieldManagement/settings/CategoryTable.jsx
```diff
@@ -278,6 +273,7 @@ export default function FieldSettings() {
       />
       <DataDisplay
         noTitleBar
+        idKey="name"
```
Emily-Ke, line 276: This now causes the duplicate key warning for custom categories

Emily-Ke (changes requested):

src/pages/userManagement/UserManagerCollaborationEditTable.jsx
```diff
@@ -164,9 +150,7 @@ export default function UserManagersCollaborationEditTable({
     {
       name: 'actions',
       align: 'left',
```
Emily-Ke, line 152: Since you are updating the other alignments, this should be updated too.

src/pages/userManagement/UserManagerCollaborationEditTable.jsx
Emily-Ke: That's not the behavior that I see. Revoking and restoring do update the table. While they are in progress, the entire table shows a progress bar. `collaborationLoading` doesn't change during revoke or restore, so the skeleton only shows up on the initial page load.

src/models/server/useServerStatus.js
```diff
@@ -105,7 +105,7 @@ export default function useServerStatus() {
     const fetchData = async () => {
       try {
         const result = await axios(
-          'https://kaiju.dyn.wildme.io:5005/api/engine/job/status/',
+          'https://kaiju.dyn.wildme.io:5010/api/engine/job/status/',
```
Atticus29, line 108: I feel like this should eventually come from site settings or environment variables or somewhere not hard-coded like this.

src/pages/userManagement/UserManagerCollaborationEditTable.jsx
Atticus29: This is also not the behavior that I see:
Screen Shot 2022-06-22 at 2 25 18 PM

Screen Shot 2022-06-22 at 2 25 26 PM
.

Atticus29 (changes requested):
I think we should leave isLoading for src/pages/userManagement/UserManagerCollaborationEditTable.jsx in. Otherwise, LGTM.

src/pages/userManagement/UserManagerCollaborationEditTable.jsx
Emily-Ke: On other tables, actions are right aligned

src/models/server/useServerStatus.js
Emily-Ke: It will, DEX-1178. This is just a temporary, quick fix.

Atticus29 (approved):
Lgtm

Emily-Ke (approved):

PR #392, opened 2022-06-20
DEX-1201 update useIndividualTermQuery, useSightingTermQuery, and bui…

Atticus29:
…ldMatchingSetQuery queries to accommodate the change in type to keyword.

This is the front-end part of the update; it seems to work, but please let me know if there's more to be done.

@brmscheiner this only works for complete guids. Substrings thereof don't match. This seems like something inherent to the keyword type? I don't know if this is desired or even avoidable behavior for a keyword type.

QA Notes:

With recent changes to the back end, the queries were failing with a 400 error:

Screen Shot 2022-06-20 at 2 49 00 PM

Changes to the query syntax seems to fix the issue:

Screen Shot 2022-06-20 at 2 50 11 PM

### DIFF
--- a/src/models/individual/useIndividualTermQuery.js
+++ b/src/models/individual/useIndividualTermQuery.js
@@ -6,7 +6,7 @@ export default function useIndividualTermQuery(searchTerm) {
     bool: {
       minimum_should_match: 1,
       should: [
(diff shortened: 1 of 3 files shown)


Emily-Ke (approved):
Looks good to me. If we want partial GUID search, we can investigate adding it in the future.

PR #394, opened 2022-06-21
update useGetUserSightings query type to term

Emily-Ke:
This is the front-end update for the back-end fix in WildMeOrg/houston#718. It does not have its own bug number, but is related to DEX-1201.


Atticus29 (approved):
Good catch!

PR #397, opened 2022-06-22
dex 1027 - filter out match results without annotations

Emily-Ke:
- Prevents a row from rendering for the match results if the `annotation_data` for the `/id_result` response does not include the annotation.
  - hot spotter includes results for annotations that houston has filtered out because they are not associated with an encounter.


Atticus29 (approved):
LGTM and looks to actually solve the problem on the UI side as well.

PR #398, opened 2022-06-23
DEX-1000 replace isSuccess with success and utilize clearSuccess to r…

Atticus29:
…emove success state upon dialog dismissal

QA notes:

Before, when you tried to assign the second encounter of a sighting to an individual, it was all:
Screen Shot 2022-06-23 at 10 47 58 AM

But, now, it be like:

Screen Shot 2022-06-23 at 10 48 34 AM


Emily-Ke (approved):
Looks like it fixes the bug.

PR #399, opened 2022-06-23
dex 1121 - pipeline status improvements

Emily-Ke:
- Redefines a "complete" status so that "failed" is no longer a sub-state of "complete"
- Internationalizes "View Match Results" button.
- Adds an alert within the timeline step if the step has failed.
error-message

This is related to WildMeOrg/houston#693


Atticus29 (approved):
Did not fully test, but cursorily tested and the code looks good!

PR #400, opened 2022-06-23
DEX-1030 and DEX-1029 change/add sortNames to firstName in individual…

Atticus29:
… and locationId in sighting

QA notes:

Individual search page should now be sortable by first name:
Screen Shot 2022-06-23 at 3 51 12 PM

And sighting search page should now be sortable by region:
Screen Shot 2022-06-23 at 3 51 35 PM

### DIFF
--- a/src/components/dataDisplays/IndividualsDisplay.jsx
+++ b/src/components/dataDisplays/IndividualsDisplay.jsx
@@ -20,9 +20,9 @@ export default function IndividualsDisplay({
   const columns = [
     {
       name: 'firstName',
-      sortName: 'firstName',
+      sortName: 'elasticsearch.firstName_keyword',
       labelId: 'NAME',
-      sortable: false,
+      sortable: true,
       options: {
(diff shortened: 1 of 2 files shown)


Emily-Ke:
The individuals sort doesn't look like it works when the individual doesn't have a name because we show "Unnamed individual" which sorts differently than an empty string.

Emily-Ke:
> The individuals sort doesn't look like it works when the individual doesn't have a name because we show "Unnamed individual" which sorts differently than an empty string.

Of course in different languages, this will fall in different places in the sort.

I'm not sure if it is the best way to handle this, but Elasticsearch supports providing a custom sorting script.
https://www.elastic.co/guide/en/elasticsearch/reference/current/sort-search-results.html#script-based-sorting

Emily-Ke:
Also, just as a QA note,  I had to create a new sighting and a new individual and then force update my Elasticsearch documents before these worked.

PR #407, opened 2022-06-28
Eslint fixes

brmscheiner:


src/pages/sighting/identification/CustomMatchingSetForm.jsx
```diff
@@ -31,7 +31,13 @@ export default function CustomMatchingSetForm({
       algorithms,
       matching_set: buildMatchingSetQuery(regionSchema, region),
     });
-  }, [algorithms?.length, region]);
+  }, [
+    setIdConfig,
```
brmscheiner, line 35: this is the one. it's because `setIdConfig` is a setstate passed in from the parent function, so the linter's not aware of it

src/components/cards/GalleryCard.jsx
```diff
@@ -20,7 +20,7 @@ export default function GalleryCard({
     <Card
       title={title}
       titleId={titleId}
-      renderActions={<Link>See all</Link>}
+      renderActions={<Link href="temp.url">See all</Link>}
```
brmscheiner, line 23: this components are not actually used, so for now i'm just making eslint shut up. soon we need to evaluate whether this component should be thrown out or if there is some valuable work to salvage.

src/constants/fieldTypesNew.js
```diff
@@ -100,14 +100,14 @@ export const fieldTypeInfo = {
...
-    exampleValue: null,
-  },
+  // [fieldTypes.file]: {
+  //   labelId: 'FILE_UPLOADER',
+  //   value: fieldTypes.file,
+  //   backendType: backendTypes.string,
+  //   backendMultiple: false,
+  //   canProvideDefaultValue: false,
+  //   exampleValue: null,
+  // },
```
brmscheiner, line 110: this is broken and without backend support. commenting out for now since tanya says we will be working on this feature soon

src/pages/individual/Individual.jsx
```diff
@@ -72,9 +72,12 @@ export default function Individual() {
             src: annotation?.asset_src,
             guid: annotation?.asset_guid,
             altText: annotation?.created
-              ? intl.formatMessage({
-                  id: 'ANNOTATION_CREATED',
-                }) + annotation?.created
+              ? [
```
Atticus29, line 75: Clever

src/pages/sighting/identification/CustomMatchingSetForm.jsx
Atticus29: Is there harm to including it?

Atticus29: That way, we could avoid the cognitive labor of repeatedly having to figure out that we should ignore this particular warning.

src/utils/notificationUtils.js
```diff
@@ -1,10 +1,8 @@
 import { get } from 'lodash-es';
-import { useIntl } from 'react-intl';
 
 import { formatDateCustom } from './formatters';
 
-export const getNotificationProps = notification => {
-  const intl = useIntl();
+export const getNotificationProps = (intl, notification) => {
```
Atticus29, line 5: Thank you!

Atticus29 (approved):
Approving this with the understanding that Ben is currently going through these and confirming that things did not break.

src/utils/notificationUtils.js
Emily-Ke: I think this fixes DEX-1113.

src/components/dataDisplays/cellRenderers/OverflowController.jsx
```diff
@@ -20,7 +20,7 @@ export default function OverflowController({ title, children }) {
   }, []);
 
   useEffect(() => {
-    if (!refEl) return;
+    if (!refEl) return Function.prototype;
```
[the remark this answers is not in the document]
Emily-Ke: This `useEffect` returns `undefined` if there is no `refEl`, but a cleanup function if there is. This breaks the `consistent-return` rule:
```
/codex-frontend/src/components/dataDisplays/cellRenderers/OverflowController.jsx
  34:5  error  Arrow function expected no return value  consistent-return
  ```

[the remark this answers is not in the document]
Comment (Emily-Ke) [reply src/components/dataDisplays/cellRenderers/OverflowController.jsx:23]: Apparently the rule is slightly more nuanced than that because `if (!refEl) return undefined;` also silences it. That might be a preferable solution.

File: src/models/encounter/useEncounterFieldSchemas.js [hunk 7]
```diff
@@ -15,9 +15,10 @@ export default function useSightingFieldSchemas() {
   const intl = useIntl();
   const { data, loading, error, siteSettingsVersion } =
     useSiteSettings();
-  if (loading || error) return null;
 
   const encounterFieldSchemas = useMemo(() => {
+    if (loading || error) return null;
```
Emily-Ke: In `useOptions` instead of exiting early from useMemo, you used the pattern:
```
  const {data, loading, error} = useCustomHook();
  const formattedData = useMemo(() => {
    // logic to format the data
  }, [data]);
  
  if (loading || error) return someDefaultValue;
  return formattedData;
```

I can think of arguments for either one, but I think we should be consistent.

src/pages/createIndividual/CreateIndividual.jsx
```diff
@@ -45,7 +45,7 @@ export default function CreateIndividual() {
   useEffect(() => {
     const initialState = calculateInitialState(createFieldSchemas);
     setFormState(initialState);
-  }, [createFieldSchemas?.length]);
+  }, [createFieldSchemas, createFieldSchemas?.length]);
```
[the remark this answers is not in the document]
Emily-Ke: You don't need the `createFieldSchemas?.length` as a dependency. `createFieldSchemas` is redeclared on every render, so this useEffect will run on every render.

src/pages/generalSettings/GeneralSettings.jsx
```diff
@@ -109,12 +109,12 @@ export default function GeneralSettings() {
...
-  const edmValues = allSettingsFields.map(fieldKey =>
-    get(siteSettings, ['data', fieldKey, 'value']),
-  );
   useEffect(() => {
+    const edmValues = allSettingsFields.map(fieldKey =>
+      get(siteSettings, ['data', fieldKey, 'value']),
+    );
     setCurrentValues(zipObject(allSettingsFields, edmValues));
-  }, edmValues);
+  }, [allSettingsFields]);
```
Emily-Ke, line 117: This hook now also has a dependency on `siteSettings`

src/pages/setup/SiteSetupForm.jsx
```diff
@@ -39,12 +39,12 @@ export default function SiteSettings({ primaryButtonId }) {
...
-  const edmValues = newSettingFields.map(fieldKey =>
-    get(newSiteSettings, ['data', fieldKey, 'value']),
-  );
   useEffect(() => {
+    const edmValues = newSettingFields.map(fieldKey =>
+      get(newSiteSettings, ['data', fieldKey, 'value']),
+    );
     setCurrentValues(zipObject(newSettingFields, edmValues));
-  }, edmValues);
+  }, [newSettingFields]);
```
Emily-Ke, line 47: This `useEffect` now has a dependency on `newSiteSettings`.
Also, `newSettingFields` does not need to be a dependency because it is defined outside of the component.

src/pages/splashSettings/SplashSettings.jsx
```diff
@@ -62,12 +62,12 @@ export default function SiteSettings() {
...
-  const edmValues = newSettingFields.map(fieldKey =>
-    get(siteSettings, ['data', fieldKey, 'value']),
-  );
   useEffect(() => {
+    const edmValues = newSettingFields.map(fieldKey =>
+      get(siteSettings, ['data', fieldKey, 'value']),
+    );
     setCurrentValues(zipObject(newSettingFields, edmValues));
-  }, edmValues);
+  }, [newSettingFields]);
```
Emily-Ke, line 70: This dependency should be `siteSettings`, not `newSettingFields`

src/pages/sighting/SightingHistoryDialog.jsx
```diff
@@ -171,7 +171,9 @@ export default function SightingHistoryDialog({ open, onClose }) {
                   <ListItemText
                     primary={
                       <span>
-                        <Link>{change.user}</Link>
+                        <Link href="temporary.url">
+                          {change.user}
+                        </Link>
```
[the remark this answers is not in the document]
Emily-Ke: Would it be better to just comment this link out too?

src/pages/sighting/identification/CustomMatchingSetForm.jsx
brmscheiner: just checked, there doesn't seem to be any harm in including it

src/pages/sighting/SightingHistoryDialog.jsx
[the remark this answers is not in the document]
brmscheiner: i have no preference. sure?

src/models/encounter/useEncounterFieldSchemas.js
```diff
@@ -94,6 +95,6 @@ export default function useSightingFieldSchemas() {
       }),
       ...customFieldSchemas,
     ];
-  }, [siteSettingsVersion]);
+  }, [intl, data, loading, error, siteSettingsVersion]);
```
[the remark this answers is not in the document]
brmscheiner: yeah, i mean theoretically if intl updated itself it would re-execute and redo the translations. i'm not sure what would happen if the user switched languages, maybe this would rerun? my bet is it just never changes though

src/pages/createIndividual/CreateIndividual.jsx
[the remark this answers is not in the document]
brmscheiner: i actually modified `createFieldSchemas` to be inside a `useMemo`. otherwise there was an infinite loop. but i am gonna switch the length check from createFieldSchemas in the useEffect to fieldSchemas on useMemo

src/utils/notificationUtils.js
brmscheiner: neat!

src/models/encounter/useEncounterFieldSchemas.js
brmscheiner: hmm, i'll update useOptions. i think short-circuiting inside the hook prevents all the code below it from running an extra time.

PR #408, opened 2022-06-29
eslint fixes - no-unsafe-optional-chaining

Emily-Ke:
Pull request checklist:

-----

- Internationalization: I think I internationalized everything in `Paginator.jsx` because it is a small module. I only internationalized the text in `Individuals.jsx` that was closely related to the `no-unsafe-optional-chaining` errors.
- Linter errors: I left some linter errors in `Individuals.jsx` because they are being addressed in #407.

### DIFF
--- a/src/pages/individual/Individual.jsx
+++ b/src/pages/individual/Individual.jsx
@@ -130,8 +130,13 @@ export default function Individual() {
             guid: annotation?.asset_guid,
             src: annotation?.asset_src,
             alt: annotation?.created
-              ? 'Annotation created: ' + annotation?.created
-              : 'Annotation image with no creation date',
+              ? intl.formatMessage(
+                  { id: 'ANNOTATION_CREATED_ON_DATE' },
+                  { date: annotation.created },
+                )
+              : intl.formatMessage({
+                  id: 'ANNOTATION_WITH_CREATION_DATE_UNKNOWN',
+                }),
           }),
         );
         return [...memo, ...modifiedAssets];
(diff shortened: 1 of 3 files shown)


src/pages/individual/Individual.jsx
```diff
@@ -72,9 +72,10 @@ export default function Individual() {
             src: annotation?.asset_src,
             guid: annotation?.asset_guid,
             altText: annotation?.created
-              ? intl.formatMessage({
-                  id: 'ANNOTATION_CREATED',
-                }) + annotation?.created
+              ? intl.formatMessage(
+                  { id: 'ANNOTATION_CREATED_ON_DATE' },
+                  { date: annotation.created },
+                )
```
Atticus29, line 78: Thank you! Much better idea.

src/pages/individual/Individual.jsx
```diff
@@ -143,7 +149,7 @@ export default function Individual() {
     );
     const firstNineAssets = slice(uniqueModifiedAssets, 0, 9);
     return firstNineAssets;
-  }, [individualData]);
+  }, [individualData, intl]);
```
Atticus29, line 152: I hadn't thought that intl would ever change?

Atticus29 (approved):
LGTM. A question, but otherwise makes sense!

src/pages/individual/Individual.jsx
Atticus29: (maybe this is a lifecycle question ultimately).

Emily-Ke: If we ever allow language switching in the application, it will.

PR #412, opened 2022-06-30
DEX-1270 remove id from optionEditor

Atticus29:
Pull request checklist:
     - N/A
     - N/A

- [ ] Thanks for keeping it clean! Feel free to merge your own pull request after it has been approved - just use the "squash & merge" option.

This PR removes the unneccesary id attribute from custom filed options. This brings newly-created custom field options in line with the current behavior of migrated custom fields. 
The removal of id here does not seem to affect the metadata cards for sightings or encounters.

It was decided that, because the end users will be exporting, bulk importing, and otherwise using these custom field options, they should be in control of what both the labels and values are (provided that these are unique - see DEX-1270 for that issue).

# QA notes:

## Old behavior when an admin tried to remove any multi-select custom field option:

Before deleting "Standing":
Screen Shot 2022-06-30 at 12 37 44 PM

After deleting "Standing":
Screen Shot 2022-06-30 at 12 34 14 PM
(i.e., they would all be removed)

## New behavior after this PR gets merged:

Before deleting "Standing":
Screen Shot 2022-06-30 at 12 37 44 PM

After deleting "Standing":
Screen Shot 2022-06-30 at 12 37 52 PM

Other things to confirm:
- [ ] Creating a new multi-select custom field and creating options works and persists without error
- [ ] Editing (add/removing) options from the above works and persists without error
- [ ] Creating a new single-select (a.k.a dropdown) custom field and creating options works and persists without error
- [ ] Editing (add/removing) options from the above works and persists without error
- [ ] Confirm that metadata cards for sightings or encounters on some old sightings page are not wonky as a result of this.

### DIFF
--- a/src/components/Button.jsx
+++ b/src/components/Button.jsx
@@ -137,46 +137,18 @@ const Core = function (
 };
 
 const ButtonWithTooltip = function (
-  {
-    showTooltip = false,
-    tooltipText = '',
-    display = 'panel',
-    loading = false,
-    style,
-    disabled,
-    size,
-    ...rest
-  },
+  { showTooltip = false, tooltipText = '', ...rest },
   ref,
 ) {
   if (showTooltip)
     return (
       <Tooltip title={tooltipText}>
         <span>
-          <CoreForwardRef
-            display={display}
-            loading={loading}
-            style={style}
-            disabled={disabled}
-            size={size}
-            ref={ref}
-            {...rest}
-          />
+          <CoreForwardRef ref={ref} {...rest} />
         </span>
       </Tooltip>
     );
-  else
-    return (
-      <CoreForwardRef
-        display={display}
-        loading={loading}
-        style={style}
-        disabled={disabled}
-        size={size}
-        ref={ref}
-        {...rest}
-      />
-    );
+  else return <CoreForwardRef ref={ref} {...rest} />;
 };
 
 const CoreForwardRef = forwardRef(Core);
--- a/src/components/Button.jsx
+++ b/src/components/Button.jsx
@@ -138,12 +138,12 @@ const Core = function (
 
 const ButtonWithTooltip = function (
   {
+    showTooltip = false,
+    tooltipText = '',
     display = 'panel',
     loading = false,
     style,
     disabled,
-    showTooltip = false,
-    tooltipText = '',
     size,
     ...rest
   },
--- a/src/components/Button.jsx
+++ b/src/components/Button.jsx
@@ -2,6 +2,7 @@ import React, { forwardRef } from 'react';
 import { FormattedMessage } from 'react-intl';
 import { useTheme } from '@material-ui/core/styles';
 import Button from '@material-ui/core/Button';
+import Tooltip from '@material-ui/core/Tooltip';
 import CircularProgress from '@material-ui/core/CircularProgress';
 import BackIcon from '@material-ui/icons/KeyboardBackspace';
 
@@ -12,6 +13,7 @@ const Core = function (
     loading = false,
     style,
     disabled,
+    tooltiptext = '',
     size,
     ...rest
   },
@@ -28,19 +30,23 @@ const Core = function (
 
   if (display === 'back') {
     return (
-      <Button
-        size="small"
-        startIcon={<BackIcon />}
-        disabled={disabled}
-        style={{
-          padding: '4px 16px',
-          ...style,
-        }}
-        ref={ref}
-        {...rest}
-      >
-        {children}
-      </Button>
+      <Tooltip title={tooltiptext}>
+        <span>
+          <Button
+            size="small"
+            startIcon={<BackIcon />}
+            disabled={disabled}
+            style={{
+              padding: '4px 16px',
+              ...style,
+            }}
+            ref={ref}
+            {...rest}
+          >
+            {children}
+          </Button>
+        </span>
+      </Tooltip>
     );
   }
 
@@ -97,18 +103,22 @@ const Core = function (
       fontSize: '14px',
     };
     return (
-      <button
-        type="button"
-        style={{ ...roleStyles, ...style }}
-        ref={ref}
-        {...rest}
-      >
-        {loading ? (
-          <CircularProgress size={24} style={spinnerStyles} />
-        ) : (
-          children
-        )}
-      </button>
+      <Tooltip title={tooltiptext}>
+        <span>
+          <button
+            type="button"
+            style={{ ...roleStyles, ...style }}
+            ref={ref}
+            {...rest}
+          >
+            {loading ? (
+              <CircularProgress size={24} style={spinnerStyles} />
+            ) : (
+              children
+            )}
+          </button>
+        </span>
+      </Tooltip>
     );
   }
 
@@ -118,21 +128,25 @@ const Core = function (
   }
 
   return (
-    <Button
-      color={color}
-      variant={variant}
-      disabled={disabled}
-      style={{ ...roleStyles, ...style }}
-      size={size}
-      ref={ref}
-      {...rest}
-    >
-      {loading ? (
-        <CircularProgress size={24} style={spinnerStyles} />
-      ) : (
-        children
-      )}
-    </Button>
+    <Tooltip title={tooltiptext}>
+      <span>
+        <Button
+          color={color}
+          variant={variant}
+          disabled={disabled}
+          style={{ ...roleStyles, ...style }}
+          size={size}
+          ref={ref}
+          {...rest}
+        >
+          {loading ? (
+            <CircularProgress size={24} style={spinnerStyles} />
+          ) : (
+            children
+          )}
+        </Button>
+      </span>
+    </Tooltip>
   );
 };
 
--- a/src/components/Button.jsx
+++ b/src/components/Button.jsx
@@ -13,7 +13,6 @@ const Core = function (
     loading = false,
     style,
     disabled,
-    tooltiptext = '',
     size,
     ...rest
   },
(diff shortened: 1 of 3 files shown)


Emily-Ke (approved):
Looks good to me.

src/pages/fieldManagement/settings/saveField/OptionEditor.jsx
```diff
@@ -42,7 +41,7 @@ export default function OptionEditor({
                 flexDirection: 'column',
                 marginBottom: 12,
               }}
-              key={option.id}
+              key={option.value}
```
Emily-Ke, line 44: Option values and/or labels cannot be guaranteed to be unique within this component. If a user clicks "Add Another Option" several times, each option looks exactly the same: `[{ label: '', value: '' }, { label: '', value: '' }, // etc.]`

Emily-Ke (dismissed):

src/pages/fieldManagement/settings/saveField/OptionEditor.jsx
Emily-Ke: Also, "value" is not a stable identifier for a key because whenever an option's value is updated, the key changes. This causes the value input to lose focus.

Atticus29: Hm...what would you recommend in this case? Just generate a uuid?

Atticus29: Option values and/or labels cannot be guaranteed to be unique within this component. If a user clicks "Add Another Option" several times, each option looks exactly the same: [{ label: '', value: '' }, { label: '', value: '' }, // etc.]

I think that this will at least be partially ameliorated by Jon's back end work on DEX-1270 (this makes me realize that I mis-typed and this whole FE PR is for ticket DEX-1120).

Emily-Ke: It will be partially ameliorated because we will be able to know that the values from the database are unique. However, while values are being added and edited within the dialog, it will always be possible for multiple value fields to be the same.

brmscheiner: once the values from the database are unique, it might be sufficient to add a timestamp property to objects created on the frontend ie. `{ label, value, created: Date.now() }`. obviously the backend objects wouldn't have these but you could use `key={o?.created || o?.value}`

brmscheiner: > Also, "value" is not a stable identifier for a key because whenever an option's value is updated, the key changes. This causes the value input to lose focus. 
for this to be resolved we would need to make sure the value used in the keys was the backend value, not the frontend's input value that changes as typed

Atticus29: Ok... with the most recent changes, I believe that the only way that two options would have the same key would be if 1) they were migrated and 2) they had the same exact value. 

I would argue that the behavior of deletion of any one of these resulting in the deletion of all of them is a feature rather than a bug, because it helps improve the usability of the product.

src/components/Button.jsx
```diff
@@ -28,19 +30,23 @@ const Core = function (
...
-        style={{
-          padding: '4px 16px',
-          ...style,
-        }}
-        ref={ref}
-        {...rest}
-      >
-        {children}
-      </Button>
+      <Tooltip title={tooltiptext}>
```
brmscheiner, line 33: I know this makes things more complicated, but the Tooltip component shouldn't be rendered unless it is being used

src/pages/fieldManagement/settings/saveField/OptionEditor.jsx
```diff
@@ -110,7 +117,24 @@ export default function OptionEditor({
...
       </DialogContent>
       <DialogActions style={{ padding: '0px 24px 24px 24px' }}>
-        <Button display="primary" onClick={onSubmit}>
+        <Button
+          disabled={
+            displayedOptions.filter(
+              option => !option?.value || !option?.label,
+            ).length > 0
+          }
+          tooltiptext={
```
brmscheiner, line 126: should be camelCase (tooltipText)

brmscheiner (approved):

src/components/Button.jsx
```diff
@@ -136,22 +136,67 @@ const Core = function (
...
+const ButtonWithTooltip = function (
+  {
+    display = 'panel',
+    loading = false,
+    style,
+    disabled,
+    showTooltip = false,
+    tooltipText = '',
+    size,
+    ...rest
```
Emily-Ke, line 148: Anything that is being passed through can just be part of `...rest`. I think you only need:
```
{    
  showTooltip = false,
  tooltipText = '',
  ...rest
}
```

src/pages/fieldManagement/settings/saveField/OptionEditor.jsx
```diff
@@ -112,7 +117,25 @@ export default function OptionEditor({
...
+          showTooltip
+          tooltipText={
+            displayedOptions.filter(
+              option => !option?.value || !option?.label,
+            ).length > 0
+              ? intl.formatMessage({
+                  id: 'UNFINISHED_OPTIONS',
+                })
+              : ''
+          }
```
Emily-Ke, line 135: When the options are valid, there is no need for a Tool tip.

src/pages/fieldManagement/settings/saveField/OptionEditor.jsx
```diff
@@ -112,7 +117,25 @@ export default function OptionEditor({
...
+          disabled={
+            displayedOptions.filter(
+              option => !option?.value || !option?.label,
+            ).length > 0
+          }
+          showTooltip
+          tooltipText={
+            displayedOptions.filter(
+              option => !option?.value || !option?.label,
+            ).length > 0
```
Emily-Ke, line 130: It might be worth defining an `areOptionsValid` variable for clarity.

src/pages/fieldManagement/settings/saveField/OptionEditor.jsx
```diff
@@ -42,7 +47,7 @@ export default function OptionEditor({
                 flexDirection: 'column',
                 marginBottom: 12,
               }}
-              key={option.id}
+              key={option?.created || option?.value}
```
Emily-Ke, line 50: `value` is still not a reliable key.
1. As you edit a previously saved option's value, this key changes on every render, which causes the input to loose focus.
2. Because you are editing values, it is still possible for two previously saved options to have the same value.

Emily-Ke (changes requested):

src/components/Button.jsx
Atticus29: Removing everything else actually broke things. Curious.

src/pages/fieldManagement/settings/saveField/OptionEditor.jsx
Atticus29: Changed in https://github.com/WildMeOrg/codex-frontend/pull/412/commits/b2c0f77857f9e4d41298476ae26b4473b092be20

src/pages/fieldManagement/settings/saveField/OptionEditor.jsx
Atticus29: Changed in https://github.com/WildMeOrg/codex-frontend/pull/412/commits/b2c0f77857f9e4d41298476ae26b4473b092be20

src/components/Button.jsx
Emily-Ke: I didn't see any errors with this:
```
const ButtonWithTooltip = function (
  {
    showTooltip = false,
    tooltipText = '',
    ...rest
  },
  ref,
) {
  if (showTooltip)
    return (
      <Tooltip title={tooltipText}>
        <span>
          <CoreForwardRef ref={ref} {...rest} />
        </span>
      </Tooltip>
    );
  else
    return (
      <CoreForwardRef ref={ref} {...rest} />
    );
};
```

Atticus29: 🤦 Thank you. Addressed in https://github.com/WildMeOrg/codex-frontend/pull/412/commits/75e6ad4191a8ebe4465dd63000d61fa5ddf19a09

Emily-Ke (approved):

src/pages/fieldManagement/settings/saveField/OptionEditor.jsx
```diff
@@ -30,66 +44,72 @@ export default function OptionEditor({
...
+                    />
+                  </FormControl>
+                </div>
               </div>
-            </div>
-          );
-        })}
+            );
+          },
+          [],
```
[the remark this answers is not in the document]
Atticus29: WHY DO I KEEP DOING THIS?

src/pages/fieldManagement/settings/saveField/OptionEditor.jsx
```diff
@@ -19,9 +20,22 @@ export default function OptionEditor({
...
   const options = value || [];
   const displayedOptions =
-    options.length > 0 ? options : [{ label: '', value: '', id: 6 }];
+    options.length > 0 ? options : [{ label: '', value: '' }];
+
+  const areAllOptionsNonEmpty =
+    filter(
+      displayedOptions,
+      option => !option?.value || !option?.label,
+    ).length === 0;
```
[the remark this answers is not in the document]
Atticus29: Good call

src/pages/fieldManagement/settings/saveField/OptionEditor.jsx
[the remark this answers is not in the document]
Atticus29: Fix in https://github.com/WildMeOrg/codex-frontend/pull/412/commits/0f31a48b6798b56fe0d29a2d5fdfd9f67ab866e3

src/pages/fieldManagement/settings/saveField/OptionEditor.jsx
```diff
@@ -19,9 +20,22 @@ export default function OptionEditor({
...
+  const areAllValuesUnique =
+    uniq(map(displayedOptions, option => option?.value)).length ===
```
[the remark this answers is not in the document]
Atticus29: Fix in https://github.com/WildMeOrg/codex-frontend/pull/412/commits/0f31a48b6798b56fe0d29a2d5fdfd9f67ab866e3

src/pages/fieldManagement/settings/saveField/OptionEditor.jsx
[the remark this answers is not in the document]
Atticus29: Fix in https://github.com/WildMeOrg/codex-frontend/pull/412/commits/0f31a48b6798b56fe0d29a2d5fdfd9f67ab866e3

PR #435, opened 2022-07-28
922 follow up

Atticus29:
Pull request checklist:

    - N/A
    - N/A
    - N/A

DEX-922

This follows up on JH's QA feedback that the Popper was buried beneath the stickyHeader.

After a lot of unnecessary other steps, I ended up just ramping up the z-index of the Popper. I predict that this is not best practice, but am happy to be wrong about this. 

I don't know what the alternative could be for Popper. Due to aforementioned unnecessary other steps, I happen to have a [minimal reproducible example](https://stackblitz.com/edit/react-wzniay?file=demo.tsx) handy in case we want to collectively ask SO about best practice with this one. Seems like overkill to me :shrug:.

### DIFF
--- a/src/components/dataDisplays/DataDisplay.jsx
+++ b/src/components/dataDisplays/DataDisplay.jsx
@@ -146,8 +146,7 @@ export default function DataDisplay({
         anchorEl={anchorEl}
         placement="bottom-end"
         transition
-        disablePortal
-        // style={{ zIndex: 9999 }}
+        style={{ zIndex: 9999 }}
       >
         {({ TransitionProps }) => (
           <Fade {...TransitionProps} timeout={350}>
--- a/src/components/dataDisplays/DataDisplay.jsx
+++ b/src/components/dataDisplays/DataDisplay.jsx
@@ -146,6 +146,8 @@ export default function DataDisplay({
         anchorEl={anchorEl}
         placement="bottom-end"
         transition
+        disablePortal
+        // style={{ zIndex: 9999 }}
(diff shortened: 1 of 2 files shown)


src/components/dataDisplays/DataDisplay.jsx
```diff
@@ -144,6 +144,7 @@ export default function DataDisplay({
         anchorEl={anchorEl}
         placement="bottom-end"
         transition
+        style={{ zIndex: 9999 }}
```
Emily-Ke, line 147: Raising the `zIndex` seems like an acceptable solution. I think that 9999 is too high though. Currently it displays on top of the app header:
overlap

The app header's z index is `theme.zIndex.drawer + 1`
https://github.com/WildMeOrg/codex-frontend/blob/17a6cb5b0c9e52a97b6e3da7c51f2c8e8419c33c/src/components/AuthenticatedAppHeader/index.js#L60

Maybe for now a zIndex of `theme.zIndex.appBar - 1` would be good enough?

Emily-Ke: It would be interesting to see if we could leverage the stacking context at some point to make sure that the filter popover is always on top within the data display, but otherwise isn't stacking on top of other components because of its z index.

https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Positioning/Understanding_z_index/The_stacking_context

Emily-Ke (approved):
Looks good to me.

PR #441, opened 2022-08-11
1379 site setting FE changed

hwindsor:
OK I need to learn a bit more react. 
The put changes work and are simplified. 
The useSiteSertting2 is intentionally bonkers as I don't have the react knowledge to do this properly.
I tested this by changing one useage of the  userSiteSettings to my new one and while it seemed to show the same data, it also blew up due to a recursion limit 
Any help on that gratefully received. 
I presume something along these lines but also needs file changes to use the common api if that happens.


src/models/site/useSiteSettings2.js
```diff
@@ -0,0 +1,45 @@
...
+import { get, merge } from 'lodash-es';
+import axios from 'axios';
+import { useQuery } from 'react-query';
+
+import queryKeys from '../../constants/queryKeys';
+
+export default function useSiteSettings() {
+
+  const settingsResult = useQuery(
+    queryKeys.settingsSchema,  // Howard so what does this actually achieve? it seems to be a text string. Is it for debug or something?
```
brmscheiner, line 11: The purpose of a query key is to give us a way to invalidate cache and trigger refreshes from other places in the app. So for example in usePostSettingsAsset on line 27 you can see us using this same string to invalidate the cache after a new image has been added, which causes another GET of the settings config.

src/models/site/useSiteSettings2.js
```diff
@@ -0,0 +1,45 @@
...
+
+
+  // Howard If the FE actually uses this for anything, we need to know as it's no longer populated with anything sane now EDM has gone
+  // (and it wasn't particularly sane beforehand) 
+  const siteSettingsVersion = {};
+
+  // Howard Regard this basically as prototype code. I wanted to prove that this was the same as the 'old parsed data' and I think it is
+    
+  console.log('new parsed data', settingData , siteSettingsVersion, );
+  return useMemo(
```
brmscheiner, line 40: there were a lot of complicated things about the old implementation of this, due to our need to combine the two requests. it should be possible to make the new version very simple, something like this:

```
import useGet from '../../hooks/useGet';
import queryKeys from '../../constants/queryKeys';

export default function useGetMe() {
  return useGet({
    queryKey: queryKeys.settingsConfig,
    url: '/site-settings/data',
    queryOptions: {
      refetchOnMount: false,
      refetchOnWindowFocus: false,
    },
  });
}

```

src/models/site/useSiteSettings2.js
brmscheiner: Also just wanted to mention, for the old implementation there were two query keys since there were two sources of data. Since we are going down to one source of data we should be able to simplify to just one query key

src/models/site/useSiteSettings2.js
hwindsor: I changed useGet to useFetch as useGet didn't exist but still getting the error.

brmscheiner:
closing in favor of #449

PR #444, opened 2022-08-15
dex 1017 - individual photo gallery

Emily-Ke:
Pull request checklist:

  - DEX-1017 - Individual photo gallery

## Pull Request Overview

- Adds the gallery page for an individual

---

**Review Notes**
- In order to avoid requesting a temporary API change from Houston that would add the asset's dimensions to the GET individual response, I derived the annotations for a gallery image based on the rendered image dimensions. This means that any image with a raw length or height greater than the master Houston size of 4096 x 4096 cannot have its annotations derived because the received image's dimensions have been clamped and are not in sync with the annotation bounds. It also means that the images requested from Houston are larger than they currently would need to be if only the asset without its annotations was being displayed.


src/pages/individualGallery/IndividualGallery.jsx
```diff
@@ -0,0 +1,166 @@
...
+import LoadingScreen from '../../components/LoadingScreen';
+import MainColumn from '../../components/MainColumn';
+import SadScreen from '../../components/SadScreen';
+import Text from '../../components/Text';
+import GalleryItem from './components/GalleryItem';
+
+function transformToAssets(individualData) {
+  if (!individualData?.encounters) return [];
+
+  const dataByAsset = transform(
```
brmscheiner, line 24: I don't love transform, because it allows memo to be mutated instead of returned and that often leads to buggy/messy code. Would it be ok to use reduce instead?

src/pages/individualGallery/IndividualGallery.jsx
```diff
@@ -0,0 +1,166 @@
...
+          const {
+            guid,
+            bounds,
+            ia_class: iAClass,
+          } = annotation || {};
+          const annotationData = { guid, bounds, iAClass };
+          const assetGuid = annotation?.asset_guid;
+
+          if (assetGuid) {
+            if (!memo[assetGuid]) {
```
brmscheiner, line 56: `if (assetGuid && !memo[assetGuid])`

src/pages/individualGallery/IndividualGallery.jsx
```diff
@@ -0,0 +1,166 @@
...
+              const assetMetadata = { src: annotation.asset_src };
```
brmscheiner, line 57: `annotation?.asset_src`

src/pages/individualGallery/IndividualGallery.jsx
```diff
@@ -0,0 +1,166 @@
...
+}
+
+export default function IndividualGallery() {
+  const intl = useIntl();
+  const theme = useTheme();
+  const { guid } = useParams();
+
+  const { data, statusCode, loading, error } = useIndividual(guid);
+
+  const galleryAssets = transformToAssets(data);
```
brmscheiner, line 86: thoughts on `useMemo` here? seems like it could be a moderately hefty operation

src/pages/individualGallery/IndividualGallery.jsx
```diff
@@ -0,0 +1,166 @@
...
+  const gridColumnWidth = 300;
```
brmscheiner, line 87: can define this outside of render

src/pages/individualGallery/components/DerivedAnnotatedPhotograph.jsx
```diff
@@ -0,0 +1,114 @@
...
+    background: theme.palette.grey['600'],
+  };
+
+  const errorCardStyles = {
+    ...imageCardStyles,
+    padding: 8,
+    display: 'flex',
+    justifyContent: 'center',
+    alignItems: 'center',
+  };
```
brmscheiner, line 85: could be defined outside of render

src/pages/individualGallery/components/FormattedLocationWithDialog.jsx
```diff
@@ -0,0 +1,49 @@
...
+const buttonAsLinkStyles = {
+  textTransform: 'none',
+  border: 'none',
+  padding: 0,
+  textDecoration: 'underline',
+  cursor: 'pointer',
+  backgroundColor: 'transparent',
+};
+
+export default function FormattedLocationWithDialog({
```
brmscheiner, line 15: what do you think of getting these formattedXYZ components in a subfolder?

brmscheiner (approved):
looks good to me. a few simple things to address before merging

src/pages/individualGallery/IndividualGallery.jsx
Emily-Ke: Do you prefer
```
  if (assetGuid && !memo[assetGuid]) // some code
  
  if (assetGuid && memo[assetGuid]) {
    // some other code that requires an asset guid and assumes memo[assetGuid]) already exists
  }
```
instead of
```
  if (assetGuid) {
    if (!memo[assetGuid]) {
       // some code
    }
    // some other code that requires an asset guid and assumes memo[assetGuid]) already exists
  }
```

src/pages/individualGallery/IndividualGallery.jsx
Emily-Ke: This line is only run if `annotation?.asset_guid` is true, so `annotation` will not be nullish.

src/pages/individualGallery/IndividualGallery.jsx
Emily-Ke: I didn't want to add it unless we thought that we needed it. I'll add it.

src/pages/individualGallery/components/DerivedAnnotatedPhotograph.jsx
Emily-Ke: I'm not sure about that. `imageCardStyles` depends on the `theme`, which is provided from a hook.

src/pages/individualGallery/components/FormattedLocationWithDialog.jsx
Emily-Ke: They are already in a subfolder: `pages/individualGallery/components`. Do you want them further nested within that folder, `pages/individualGallery/components/formatted`? I don't think that is necessary, but I don't mind making the change.

src/pages/individualGallery/IndividualGallery.jsx
Emily-Ke: Switched to `Array.prototype.reduce()` in [4e996b](https://github.com/WildMeOrg/codex-frontend/pull/444/commits/4e996bf7697253627fefe76ce3a6cf495d59879d)

src/pages/individualGallery/IndividualGallery.jsx
Emily-Ke: Added in https://github.com/WildMeOrg/codex-frontend/pull/444/commits/4e996bf7697253627fefe76ce3a6cf495d59879d

src/pages/individualGallery/IndividualGallery.jsx
Emily-Ke: Moved in https://github.com/WildMeOrg/codex-frontend/pull/444/commits/4e996bf7697253627fefe76ce3a6cf495d59879d

src/pages/individualGallery/IndividualGallery.jsx
brmscheiner: i think for really trivial things it doesnt make a difference. but there's enough computation here i think it is nice to add

src/pages/individualGallery/components/FormattedLocationWithDialog.jsx
brmscheiner: oh nope, didn't notice that.

src/pages/individualGallery/IndividualGallery.jsx
brmscheiner: eh, nah, this is fine

PR #448, opened 2022-08-17
Proposed table action group refactor

brmscheiner:
This proposed component would allow us to 
1. simplify creating tables with row actions
2. apply styles for these actions consistently across the board, and 
3. provide a way to implement these row actions cleanly without linter errors

thoughts?


src/pages/fieldManagement/settings/CustomFieldTable.jsx
```diff
@@ -6,8 +6,8 @@ import Grid from '@material-ui/core/Grid';
 import Tooltip from '@material-ui/core/Tooltip';
 import AddIcon from '@material-ui/icons/Add';
 
+import { cellRendererTypes } from '../../../components/dataDisplays/cellRenderers';
```
brmscheiner, line 9: just one example refactor, next step would be to do it across the board

src/pages/fieldManagement/settings/CustomFieldTable.jsx
```diff
@@ -62,6 +62,25 @@ export default function CustomFieldTable({
...
+  const deriveEditHref = useMemo(
+    () => (_, field) =>
+      `/settings/fields/save-custom-field/${fieldTypeName}/${field.id}`,
+    [fieldTypeName],
+  );
+
+  const onDeleteCustomField = useMemo(
+    () => (_, field) => setDeleteField(field),
+    [],
+  );
```
Emily-Ke, line 82: These can be simplified to `useCallback`s
> useCallback(fn, deps) is equivalent to useMemo(() => fn, deps)
https://reactjs.org/docs/hooks-reference.html#usecallback

brmscheiner: good point! i always forget about that one

brmscheiner:
@Emily-Ke changes pushed!

PR #449, opened 2022-08-17
Integrate new site settings API

brmscheiner:
Pull request checklist:

  - DEX-1379 - Use new site setting APIs on Frontend

## Pull Request Overview

- 

---

**Review Notes**
This PR migrates from the old site settings endpoints to the new ones, and updates things where necessary. Tested against houston branch `378_old_api_removal`. Some attempts to clean up code were made but not necessarily everywhere.


src/models/site/usePutSiteSettings.js
```diff
@@ -3,10 +3,10 @@ import queryKeys from '../../constants/queryKeys';
 
 export default function usePutSiteSettings() {
```
hwindsor, line 4: Any reason you chose tokeep this rather than using usePutSiteSetting with a property of ''?

src/models/site/usePutSiteSettings.js
```diff
@@ -3,10 +3,10 @@ import queryKeys from '../../constants/queryKeys';
...
   return usePost({
-    url: '/site-settings/main/',
+    url: '/site-settings/data',
     deriveData: ({ data }) => data,
```
hwindsor, line 7: This should be value: data,

src/pages/match/ImageCard.jsx
```diff
@@ -14,17 +14,17 @@ export default function ImageCard({ titleId, annotation }) {
...
   const regionChoices = useMemo(
     () =>
-      get(
-        siteSettings,
-        ['site.custom.regions', 'value', 'locationID'],
-        [],
-      ),
-    [siteSettingsVersion, siteSettings],
+      get(siteSettings, [
+        'site.custom.regions',
```
hwindsor, line 23: If you only want the one value, why are you getting the block (or does this use a cache of the block)

src/pages/splashSettings/SplashSettings.jsx
```diff
@@ -226,39 +222,22 @@ export default function SplashSettings() {
...
-                      options: choices.map(choice => choice.label),
-                    };
-                  });
-
-                  currentValues[customFieldKey] = {
-                    definitions: newFields,
-                  };
-                }
-              });
               putSiteSettings({ data: currentValues });
```
hwindsor, line 225: Rather than sending potentially 4 messages here, it would be far more efficient to add the  splashVideoPostData etc to currentValues and send them as one message

src/pages/generalSettings/GeneralSettings.jsx
```diff
@@ -297,35 +292,10 @@ export default function GeneralSettings() {
...
-                      options: choices.map(choice => choice.label),
-                    };
-                  });
-
-                  currentValues[customFieldKey] = {
-                    definitions: newFields,
-                  };
-                }
-              });
               putSiteSettings({ data: currentValues });
```
hwindsor, line 295: before this, 
if logoPostData:
    currentValues['logo'] = logoPostData
Then we can get rid of the multiple posts

hwindsor:
> @Emily-Ke I've marked this ready for review but please do not merge. We will have to coordinate merge with the related houston PR(s)

This is dependent on BE DEX 1377 which is now merged so the next step is for the FE to merge this PR when ready

src/models/site/usePutSiteSettings.js
brmscheiner: @Emily-Ke - this seemed to be working as-is, but let's give howard's suggestion a try and send data that way if it works. better to be consistent

src/pages/generalSettings/GeneralSettings.jsx
brmscheiner: assets used to need this special treatment. if they dont any more, that's great! try this, but if it works chances are the component logic can be simplified even more than just this

src/pages/match/ImageCard.jsx
brmscheiner: the cache of the block. it actually should be faster to use the site settings block that was already fetched than to fetch only the value that's needed in a separate request

src/pages/splashSettings/SplashSettings.jsx
brmscheiner: super excited about this if it works!

hwindsor: Of course it works, I wrote the code :laughing:

src/pages/generalSettings/GeneralSettings.jsx
hwindsor: This is part of this simplification. The only way assets are now 'special' is that you send a transaction id and get back a file guid, other than that, a file is now identical to any other site setting

Emily-Ke (approved):
Looks good to me.

src/models/site/useSiteSettings.js
```diff
@@ -1,66 +1,10 @@
...
-
-  return useMemo(
-    () => ({ data, loading, error, siteSettingsVersion }),
-    [data, loading, error, siteSettingsVersion],
-  );
+export default function useGetSiteSettings() {
+  return useFetch({
+    queryKey: queryKeys.settingsSchema,
+    url: '/site-settings/data',
+    dataAccessor: result => result?.data?.data,
```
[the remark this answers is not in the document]
hwindsor: Presuming this question is for @brmscheiner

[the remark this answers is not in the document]
brmscheiner: Good points, we could definitely clean those things up. Regarding which name to use, I personally like `useSiteSettings` but I know Mark likes the Get in there so feel free to break the tie and choose whichever name you like better.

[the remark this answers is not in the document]
Emily-Ke: I restored it to `useSiteSettings` because that's what we currently call it when we import it.

PR #452, opened 2022-08-26
Fix page error when no notification message

Emily-Ke:
Pull request checklist:

- [x] ~New features support all states (loading, error, etc)~ **No new features**

  - None. The bug became observable because of WildMeOrg/houston#795 which introduces new notification message types.

---

## Pull Request Overview

- On the notifications page, returns `null` instead of formatted text if there is no notification message. This prevents react-intl from throwing an uncaught error when `undefined` is passed to `id`

### DIFF
--- a/src/pages/notifications/Notifications.jsx
+++ b/src/pages/notifications/Notifications.jsx
@@ -100,58 +100,59 @@ export default function Notifications() {
                 const createdDate = notification?.created;
                 const timeSince =
                   calculatePrettyTimeElapsedSince(createdDate);
-                const notificationText = (
-                  <Text
-                    key={notification?.guid}
-                    style={{
-                      color: read
-                        ? theme.palette.text.secondary
-                        : theme.palette.text.primary,
-                    }}
-                    variant="body2"
-                  >
-                    {intl.formatMessage(
-                      {
-                        id: currentNotificationSchema?.notificationMessage,
-                      },
-                      {
-                        userName: (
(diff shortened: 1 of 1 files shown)


src/pages/notifications/Notifications.jsx
```diff
@@ -100,58 +100,59 @@ export default function Notifications() {
...
-                            </Link>
-                          </span>
-                        ),
-                        formattedDeadline,
-                      },
-                    )}
-                  </Text>
-                );
+                const notificationText =
+                  currentNotificationSchema?.notificationMessage ? (
```
Emily-Ke, line 104: Almost all of this doff is formatting caused by adding this line. Everything was indented because of this.

src/pages/notifications/Notifications.jsx
```diff
@@ -100,58 +100,59 @@ export default function Notifications() {
...
+                      style={{
+                        color: read
+                          ? theme.palette.text.secondary
+                          : theme.palette.text.primary,
+                      }}
+                      variant="body2"
+                    >
+                      {intl.formatMessage(
+                        {
+                          id: currentNotificationSchema.notificationMessage,
```
Emily-Ke, line 116: This is the one other change. I removed the optional chaining operator because the new condition above that determines if this code should be run is that `currentNotificationSchema.notificationMessage` exists.

brmscheiner (approved):

PR #453, opened 2022-08-29
dex-1401 - incorrect sighting metadata payload

Emily-Ke:
Pull request checklist:

- [x] ~All text is internationalized~ **No user facing text**
- [x] ~New features support all states (loading, error, etc)~ **No new features**

  - DEX-1401 - sighting metadata editing dialog sending incorrect payload to api

---

## Pull Request Overview

- Changes `LatLongEditor`'s state so that the inputs display the `value` passed in even when that value changes after the component has first mounted.
  - Changes `LatLongEditor`'s state handling to a reducer to more easily handle the related state.
- Updates `EditSightingMetadata` so that only the first change to the metadata dependency causes an update to the form values.

### DIFF
--- a/src/components/fields/edit/LatLongEditor.jsx
+++ b/src/components/fields/edit/LatLongEditor.jsx
@@ -1,4 +1,4 @@
-import React, { useState, useEffect, useMemo } from 'react';
+import React, { useState, useEffect, useReducer } from 'react';
 import { useIntl } from 'react-intl';
 import { get } from 'lodash-es';
 import DialogContent from '@material-ui/core/DialogContent';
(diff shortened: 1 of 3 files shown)


src/components/fields/edit/LatLongEditor.jsx
```diff
@@ -15,16 +15,73 @@ function getNumberString(n) {
...
+    latitude: getNumberString(get(coordinates, '0', '')),
+    longitude: getNumberString(get(coordinates, '1', '')),
+    mapCoordinates: null,
+    setterType: 'external',
   };
 }
 
+function reducer(state, action) {
+  switch (action.type) {
+    case 'input-latitude':
```
brmscheiner, line 29: would be good to define reducer actions as constants to prevent typos from messing things up. i'm also still not a fan of these switch statements but it's not that big a deal..

brmscheiner (approved):

PR #459, opened 2022-09-27
DEX-1443 cherry pick e8c4a2d from DEX-1443-empty-dropdown-as-array be…

Atticus29:
…cause there were commits in that branch that I did not make, so this felt like the cleaner solution

# Description

This is a companion PR with BE's [PR 810](https://github.com/WildMeOrg/houston/pull/810).

It makes the defaultValue for the customField select dropdowns null instead of an empty string.

It does not solve the remaining issue in DEX-1443, which has to do with the phrasing of the error message. IMO, this error message change should occur in the back end, because we are currently parroting out the errors that we see from them (which may change, as JVO pointed out to me, once we start internationalizing them as well).

Pull request checklist:

    - NA

DEX-1443

### DIFF
--- a/src/constants/fieldTypes.js
+++ b/src/constants/fieldTypes.js
@@ -138,7 +138,7 @@ export const fieldTypeChoices = [
   {
     labelId: 'DROPDOWN',
     value: fieldTypes.select,
-    defaultValue: '',
+    defaultValue: null,
     backendType: backendTypes.string,
     backendMultiple: false,
     configuration: [
@@ -147,7 +147,7 @@ export const fieldTypeChoices = [
         descriptionId: 'DROPDOWN_CHOICE_HELPER_TEXT',
         value: 'choices',
         type: fieldTypes.optioneditor,
-        defaultValue: [],
+        defaultValue: null,
       },
     ],
   },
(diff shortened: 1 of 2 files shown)


Emily-Ke (commented):
Does this also need to be changed in fieldTypesNew.js?
https://github.com/WildMeOrg/codex-frontend/blob/4aa18da07fa0365755b182ed1dc5f90648d57291/src/constants/fieldTypesNew.js#L120

src/constants/fieldTypes.js
```diff
@@ -147,7 +147,7 @@ export const fieldTypeChoices = [
         descriptionId: 'DROPDOWN_CHOICE_HELPER_TEXT',
         value: 'choices',
         type: fieldTypes.optioneditor,
-        defaultValue: [],
+        defaultValue: null,
```
Emily-Ke, line 150: Are you sure that this should change? Choices aren't an array any more?

Emily-Ke (approved):
Approved with a question.

src/constants/fieldTypes.js
Atticus29: Since I couldn't detect any difference in behavior on the platform either way (I imagine that it makes a difference somewhere, but I can't easily find where), I've changed it back to an array here.

PR #460, opened 2022-09-28
Dex 1328 user manager view collabs

Atticus29:
# Description

## Before
When a user-manager is viewing their own profile page, they see a collaboration table:
Screen Shot 2022-09-28 at 2 08 28 PM

When a user-manager is viewing another user's profile page, they used to not see the other user's collaboration table:

Screen Shot 2022-09-28 at 2 09 06 PM

## After

With this PR, user-manager (and only user-manager) will see a table much like the one they see in the settings → manage users page, except filtered by the target nonSelf user's collaborations (indeed, the collaborations in question are sourced from a call to `useGetUser(userId)`:

Screen Shot 2022-09-28 at 2 13 54 PM

This should not affect the user-manager's view of their own page:

Screen Shot 2022-09-28 at 2 14 19 PM

And non-data-manager users should not see it:

In this case, I'm logged in as TestUser2, viewing my collaborator TestUser1's profile page.
Screen Shot 2022-09-28 at 2 15 18 PM

The user manager buttons all work, but they don't seem to refresh the table. I consider this out-of-scope for the current ticket.

Pull request checklist:

DEX-1328

### DIFF
--- a/src/components/UserProfile.jsx
+++ b/src/components/UserProfile.jsx
@@ -5,6 +5,7 @@ import { get } from 'lodash-es';
 import { getHighestRoleLabelId } from '../utils/roleUtils';
 import useUserMetadataSchemas from '../models/users/useUserMetadataSchemas';
 import useGetUserSightings from '../models/users/useGetUserSightings';
+import useGetMe from '../models/users/useGetMe';
 import useGetUserUnprocessedAssetGroupSightings from '../models/users/useGetUserUnproccessedAssetGroupSightings';
 import { formatDate, formatUserMessage } from '../utils/formatters';
 import EntityHeader from './EntityHeader';
@@ -35,7 +36,8 @@ export default function UserProfile({
   const metadataSchemas = useUserMetadataSchemas(userId);
   const { data: agsData, loading: agsLoading } =
     useGetUserUnprocessedAssetGroupSightings(userId);
-
+  const { data: currentUserData } = useGetMe();  
+  const isUserManager = get(currentUserData, 'is_user_manager', false);
   const metadata = useMemo(() => {
     if (!userData || !metadataSchemas) return [];
     return metadataSchemas
@@ -46,7 +48,7 @@ export default function UserProfile({
         ...schema,
         value: schema?.getValue(schema, userData),
       }));
-  }, [userData, metadataSchemas]);
+  }, [userData, metadataSchemas, someoneElse]);
 
   const imageSrc = get(userData, ['profile_fileupload', 'src']);
   const imageGuid = get(userData, ['profile_fileupload', 'guid']);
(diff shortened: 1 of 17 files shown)


src/components/cards/CollaborationsCard.jsx
```diff
@@ -11,8 +11,10 @@ import DataDisplay from '../dataDisplays/DataDisplay';
 import CollaborationsDialog from './collaborations/CollaborationsDialog';
 
 export default function CollaborationsCard({
+  title,
   userId,
   htmlId = null,
+  noCollaborationsMsg,
```
Emily-Ke, line 17: Do `title` and `noCollaborationsMsg` need to be props? They never change.

src/components/UserProfile.jsx
```diff
@@ -35,6 +38,27 @@ export default function UserProfile({
...
+  const isUserManager = get(
+    currentUserData,
+    'is_user_manager',
+    false,
+  );
+  const {
+    data: nonSelfUserData,
+    loading: nonSelfUserDataLoading,
+    error: nonSelfUserDataError,
+  } = useGetUser(userId);
```
Emily-Ke, line 55: The `userId` is not guaranteed to be nonSelf. When you view your own page, the userId is your own.

src/components/UserProfile.jsx
```diff
@@ -35,6 +38,27 @@ export default function UserProfile({
...
+
+  const nonSelfCollabData = get(
+    nonSelfUserData,
+    ['collaborations'],
+    [],
+  ); // deleteMe this seems like too small a calculation to memoize
```
Emily-Ke, line 61: I agree that it doesn't need to be memoized. I think you can delete the comment.

src/components/UserProfile.jsx
Emily-Ke, line 55: You might be able to just use the `userData` and `userDataLoading` props that are already being passed in.

src/components/UserProfile.jsx
```diff
@@ -18,6 +19,8 @@ import MetadataCard from './cards/MetadataCard';
 import SightingsCard from './cards/SightingsCard';
 import CollaborationsCard from './cards/CollaborationsCard';
 import CardContainer from './cards/CardContainer';
+import UserManagersCollaborationEditTable from '../pages/userManagement/UserManagerCollaborationEditTable';
```
Emily-Ke, line 22: Because this component is no longer being used only within the user management page, it should be moved to a different directory.

src/components/cards/CollaborationsCard.jsx
Atticus29: A lot of these things are vestiges of when I was populating a normal collaboration table with a userId that was different form the current viewer. 

I kept a few of the things that I thought generalized the code in ways that might future-proof the product.

But you're right; I shouldn't plan for a use case we demonstrably no longer have.

src/components/UserProfile.jsx
Atticus29: Done

src/components/UserProfile.jsx
Atticus29: Done

src/components/UserProfile.jsx
Atticus29: Done

src/components/UserProfile.jsx
Atticus29: Ended up removing this entirely.

src/components/UserProfile.jsx
```diff
@@ -171,6 +187,23 @@ export default function UserProfile({
               userId={userId}
             />
           )}
+          {someoneElse && isUserManager && (
+            <Paper
```
Emily-Ke, line 191: For consistency, this can be a `Card` instead.

Emily-Ke (approved):

src/components/cards/collaborations/collaborationStates.js
```diff
@@ -43,7 +43,7 @@ export default {
         get(collaboration, ['otherUserData', testKey]),
       ),
     currentStateMessage:
-      'Access revoked. Only your collaboration partner can restore access.',
+      'Access revoked. Only the collaboration partner or a user manager can restore access.',
```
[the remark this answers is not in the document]
Atticus29: I agree, and I think it's out-of-scope.

src/components/UserProfile.jsx
Atticus29: Done

PR #471, opened 2022-10-14
Dex 1313 add timeout user guidance

Atticus29:
# Description

If the end user is searching for a species to add to their codex instance, historically the potentially long search time for this and unreliability of the API has been a suboptimal user experience.

This PR introduces the following features:

1. After 5 seconds, if the search status is still loading and there's no error, the user will see an indication that the search is still underway:
Screen Shot 2022-10-14 at 2 50 45 PM

2. If the search errors out in any way, instead of whichever, if any, error message that would have been issued by the response, the end user will now see:
Screen Shot 2022-10-14 at 2 51 47 PM

3. The end user will be able to search again, and either the error alert or the search status ("Still generating list...") will be dismissed.

4. The end user will not see either of these indicators if the search is faster than 5 seconds:

Screen Shot 2022-10-14 at 2 52 12 PM

5. The end user will not see either of these indicators after the search is completed and there is no error:

Screen Shot 2022-10-14 at 2 52 12 PM

6. The end user can click the search button again upon completion (successful or not) call to the ITIS api:
Screen Shot 2022-10-14 at 2 51 47 PM
OR
Screen Shot 2022-10-14 at 2 52 12 PM

## QA Notes

- In order to reproduce a long search that does not complete before the 1-min. timeout, I just searched for the letter, "S".
- In order to reproduce a shorter search that does complete before the 1-min. timeout, I searched for, "Solenopsis".

Pull request checklist:

DEX-1313

### DIFF
--- a/src/pages/fieldManagement/settings/defaultFieldComponents/SpeciesEditor.jsx
+++ b/src/pages/fieldManagement/settings/defaultFieldComponents/SpeciesEditor.jsx
@@ -64,6 +64,12 @@ export default function SpeciesEditor({
     [],
   );
 
+  const isTimeoutError =
+    searchResultsError && searchResultsError?.indexOf('Timeout') > -1;
+  const isNonTimeoutError =
+    searchResultsError &&
+    searchResultsError?.indexOf('Timeout') === -1;
+
   const tableColumns = [
     {
       name: 'scientificName',
@@ -192,12 +198,17 @@ export default function SpeciesEditor({
             id="STILL_GENERATING_LIST"
           />
         )}
-        {searchResultsError && (
+        {isTimeoutError && (
           <CustomAlert
             style={{ marginTop: 12 }}
             severity="error"
             titleId="SEARCH_TIMED_OUT_WHILE_TRYING_TO_CONNECT_TO_ITIS"
           />
+        )}
+        {isNonTimeoutError && (
+          <CustomAlert style={{ marginTop: 12 }} severity="error">
+            {searchResultsError}
+          </CustomAlert>
         )}
         <DataDisplay
           cellStyles={{ padding: '0 8px 0 12px' }}
--- a/src/pages/fieldManagement/settings/defaultFieldComponents/SpeciesEditor.jsx
+++ b/src/pages/fieldManagement/settings/defaultFieldComponents/SpeciesEditor.jsx
@@ -197,9 +197,7 @@ export default function SpeciesEditor({
             style={{ marginTop: 12 }}
             severity="error"
             titleId="SEARCH_TIMED_OUT_WHILE_TRYING_TO_CONNECT_TO_ITIS"
-          >
-            {/* {searchResultsError} */}
-          </CustomAlert>
+          />
         )}
         <DataDisplay
           cellStyles={{ padding: '0 8px 0 12px' }}
--- a/src/pages/fieldManagement/settings/defaultFieldComponents/SpeciesEditor.jsx
+++ b/src/pages/fieldManagement/settings/defaultFieldComponents/SpeciesEditor.jsx
@@ -37,7 +37,6 @@ export default function SpeciesEditor({
     loading: searchResultsLoading,
     error: searchResultsError,
     setError,
-    setLoading,
   } = useItisSearch(searchTerm);
 
   const timerRef = useRef(null);
@@ -54,9 +53,8 @@ export default function SpeciesEditor({
     if (searchResultsError) {
       setStillGeneratingDisplay(false);
       clearTimeout(timerRef.current);
-      setLoading(false);
     }
-  }, [searchResultsError, searchResultsLoading, setLoading]);
+  }, [searchResultsLoading]);
   const currentSpecies = get(formSettings, 'species', []);
   const suggestedValues = get(
     siteSettings,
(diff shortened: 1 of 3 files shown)


src/utils/useItisSearch.js
```diff
@@ -59,5 +59,5 @@ export default function useItisSearch(searchKey, maxResults = 50) {
     if (searchKey) fetchData();
   }, [searchKey, maxResults]);
 
-  return { data, loading, error };
+  return { data, loading, error, setError, setLoading };
```
Emily-Ke, line 62: What is the use case for an external component telling a request that it is or is not loading?

Emily-Ke, line 62: The only time that I see `setError` being used is to clear the error when the Search button is clicked. It probably makes more sense to clear the error within `useItisSearch` so that it is always reset before a request is made.

src/pages/fieldManagement/settings/defaultFieldComponents/SpeciesEditor.jsx
```diff
@@ -30,12 +30,33 @@ export default function SpeciesEditor({
...
+    if (!searchResultsLoading) {
+      setStillGeneratingDisplay(false);
+      clearTimeout(timerRef.current);
+    }
+    if (searchResultsError) {
+      setStillGeneratingDisplay(false);
+      clearTimeout(timerRef.current);
+      setLoading(false);
+    }
+  }, [searchResultsError, searchResultsLoading, setLoading]);
```
Emily-Ke, line 59: The message about loading taking a long time ideally should only have to depend on loading. Did loading continue for a noticeable time after an error message appeared?

src/pages/fieldManagement/settings/defaultFieldComponents/SpeciesEditor.jsx
```diff
@@ -160,10 +184,20 @@ export default function SpeciesEditor({
...
+        )}
         {searchResultsError && (
-          <CustomAlert severity="error">
-            {searchResultsError}
-          </CustomAlert>
+          <CustomAlert
+            style={{ marginTop: 12 }}
+            severity="error"
+            titleId="SEARCH_TIMED_OUT_WHILE_TRYING_TO_CONNECT_TO_ITIS"
+          />
```
Emily-Ke, line 200: ❗ We should still show the `searchResultsError` if it fails for some reason other than the timeout.

Emily-Ke (changes requested):

src/pages/fieldManagement/settings/defaultFieldComponents/SpeciesEditor.jsx
Atticus29: Yes, totally... the error would appear, and loading would just continue to stay true.

src/utils/useItisSearch.js
Atticus29: See above:
the error would appear, and loading would just continue to stay true.

Atticus29: Could you elaborate on this one? I left a question about it on slack.

src/pages/fieldManagement/settings/defaultFieldComponents/SpeciesEditor.jsx
Atticus29: Responded in https://github.com/WildMeOrg/codex-frontend/pull/471/commits/49db30f76452f80b3cdce12a2017a5600f193e96

src/utils/useItisSearch.js
Emily-Ke: That is a bug in useItisSearch. I think it should be fixed there.

src/pages/fieldManagement/settings/defaultFieldComponents/SpeciesEditor.jsx
Emily-Ke: Fixing the bug in useItisSearch should mean that you can simplify this to not depend on `searchResultsError` or `setLoading`

src/pages/fieldManagement/settings/defaultFieldComponents/SpeciesEditor.jsx
```diff
@@ -151,7 +178,10 @@ export default function SpeciesEditor({
             style={{ width: '100%' }}
           />
           <Button
-            onClick={() => setSearchTerm(searchInput)}
+            onClick={() => {
+              setError(null);
+              setSearchTerm(searchInput);
+            }}
```
Emily-Ke, line 184: Because this is the submission button for a form, `setError(null)` can be moved to the form's `onSubmit`.

src/pages/fieldManagement/settings/defaultFieldComponents/SpeciesEditor.jsx
```diff
@@ -54,9 +53,8 @@ export default function SpeciesEditor({
     if (searchResultsError) {
       setStillGeneratingDisplay(false);
       clearTimeout(timerRef.current);
```
Emily-Ke, line 55: This can be removed also. `searchResultsError` can be removed from the first condition. That means the second condition can just become and `else` block.

Emily-Ke (approved):
Looks good.

src/utils/useItisSearch.js
```diff
@@ -59,5 +63,5 @@ export default function useItisSearch(searchKey, maxResults = 50) {
...
```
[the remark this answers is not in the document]
Atticus29: Responded in https://github.com/WildMeOrg/codex-frontend/pull/471/commits/03c337351e5822d9fc53f6a8296565872350bd6e

PR #472, opened 2023-03-30
PR#846-frontend

erinz2020:
Pull request checklist:

- [x]  All text is internationalized N/A
- [x] New features support all states (loading, error, etc) N/A

This PR is for the frontend of PR#846 
Pull Request Overview
When a Bulk Import is done, the user (usually) commits the related sightings one at a time in sequence. Each commit (usually) triggers an ID job. The problem is, each ID jobs is matching against one (or more) additional Annotation than the previous, which is a nightmare for caching on Sage.

Proposed solution in this PR is to modify the matching-set query such that it will include all Annotations within the same AssetGroup so as to hopefully have a better chance of having cache-hits in Sage.

Modify Annotation Elasticsearch schema to include AssetGroup (aka GitStore) guid
Adjust matching-set query to including Annotations in same AssetGroup even when they have no Encounter (yet)

### DIFF
--- a/src/pages/sighting/identification/buildMatchingSetQuery.js
+++ b/src/pages/sighting/identification/buildMatchingSetQuery.js
@@ -40,30 +40,35 @@ export default function buildMatchingSetQuery(regionSchema, region) {
   );
 
   return {
-    bool: {
-      filter: [
+    "bool": {
+      "minimum_should_match": 1,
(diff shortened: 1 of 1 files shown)


naknomum (approved):
looks like the right changes to me, and seemed to give the right results during our testing.

PR #475, opened 2023-04-24
cdx-7 add species on edit individual metadata page

erinz2020:
1. Add a field schema "species"
2. List all site species as options, provides both the scientific and the common names
3. Add it on "edit individual metadata" page, and make it required.


naknomum (approved):
looks good to me.

PR #476, opened 2023-04-25
cdx-37 add species on individual creating page, also add custom alert…

erinz2020:
1. Add field "species" on individual creating page and make it required
2. Add custom alert for required fields


naknomum (approved):
going to have to trust that this react stuff is doing what i am _guessing_ it is doing; so going to sign off on this cuz i dont see anything off here based on that.

PR #479, opened 2023-05-16
Generalize contact information

TanyaStere42:
Contact information was outdated; needs to go to a general location rather than a specific email


naknomum (approved):
er, yup.

PR #480, opened 2023-05-18
cdx-19-20-sightings

erinz2020:
1. fix search by regions bug
2. add some standard fields for search
3. add all custom fields for search exclude type "individual" and "feet/meter selector"


naknomum (approved):
looks good

PR #481, opened 2023-05-19
cdx-21-22

erinz2020:
1. add relationship/relationship role/social group/social group role to individual search 
2. add all custom fields to individual search


naknomum (approved):
lgtm

PR #482, opened 2023-05-22
cdx-1-scores_by_individual

erinz2020:
1. display match result's score using individual score instead of annotation score
2. Insert a column 'Individual name'


naknomum (approved):
lgtm

naknomum (approved):
new changes look good

PR #491, opened 2023-06-02
cdx-8

erinz2020:
Add species as a required field when merging two individuals.
    1. If two individuals have same valid species, populate this species, user can't adjust
    2. If two individuas have different valid species, populate these two species, let user choose one
    3. If one individual has species while the other not, populate the species, user can't adjust
    4. If both individuals have no species, populate all species of this site, let user choose one


naknomum (approved):
looks good

PR #492, opened 2023-06-05
next button also saves changes

erinz2020:
1. when user clicks at next button, all changes will be saved
2. if user sets "allow twitter submissions" to true, but don't fill in required fields, next button on that page will be disabled


naknomum (approved):
seems good

PR #493, opened 2023-06-05
cdx-12 add policy confirmation

erinz2020:
On profile set up page:
1. add policy confirmation, if not checked, user won't be able to proceed
2. update: if username is whitespace, user won't be able to proceed


naknomum (approved):
look good

PR #494, opened 2023-06-08
update href of link

erinz2020:
1. update the links in terms and conditions(set up profile page), combine two links to be one
2. when click at the link, a new tab will be opened with the pdf file about TOU for Codex


naknomum (approved):
looks good

PR #496, opened 2023-06-20
Account request

erinz2020:
On Request an invitation page, 
1. when user fill in required email address and name and click SEND REQUEST button, send POST request to /api/v1/account-requests
2. at the same time send recaptcha token
3. deal with response status, if status is 200, show successful page, if error, show error message

Create a new page:  /requests/ to display requests from invitation page
1. Send GET method to BE and get response
2. display requests list on the page


naknomum (approved):
looks good

PR #497, opened 2023-07-05
cdx-15-fix-date-custom-fields-issue

erinz2020:
Fix date picker/date range custom fields won't save issue on individual/sighting/encounter metadate editing page


naknomum (approved):
looks good

PR #499, opened 2023-07-13
cdx-26-patch

erinz2020:
1. Add a translation for label "INVITATION_REQUESTS" which got lost during merging
2. On report a sighting page, move required encounter fields to 'required data' page
3. Update required fields check function


naknomum (approved):
lgtm

PR #500, opened 2023-07-14
cdx-26-patch-2

erinz2020:
when report a sighting, encounter fields should only show up when user selects "one animal"


naknomum (approved):
yup

PR #513, opened 2023-07-21
cdx-56-update-theme-color-and-text-color

erinz2020:
1. update theme primary colors
2. update black color in theme
3. add secondary and tertiary text color in theme


naknomum (approved):
looks good

PR #514, opened 2023-07-24
cdx-57-update-button-styles

erinz2020:
1. update all buttons' border radius
2. update secondary buttons' variant and text color
3. update tertiary buttons' text color to reflect theme color
4. buttons allow just text, just icons, icon before text, and text before icon


naknomum (approved):
lgtm

Issue #527, opened 2023-07-31
Autogenerate names/Codex ID - Codex ID on Individual page


TanyaStere42:
Context: arguments about names are a significant reason against collaboration within biology. People want to maintain the specific name they leverage in their personal catalogue.

By providing an automatically generated name that is associated with Codex specifically, we can promote Codex as a collaboration tool, as well as a catalogue solution, by providing a workspace with common shared nomenclature that does not threaten the name spaces used in individual catalogues.

## Resources

Final design: https://www.figma.com/file/xWzUldKxiSvoKsQO4Y7azD/Codex-ID?type=design&node-id=0-1&mode=design&t=LrVVnr6vj3lJvtbo-0

Note: design assumes https://github.com/WildMeOrg/codex-frontend/issues/508 has been completed.

![Image](https://user-images.githubusercontent.com/62726866/257299178-cf9bb431-fa7e-49bf-9f6b-1dcf3ff1d4f8.png)

User Story:

As a user on an individual page, I see the Codex ID as a chip in the summary section at the top of the individual page.

PR #540, opened 2023-08-16
add chip under on individual page

erinz2020:
1. show individual's codex ID on individual page
2. if no codex ID, just show '-'

covers issue #527


naknomum (approved):
looks good

Issue #523, opened 2023-07-24
Can't save Metadata edits on an Individual page


goddesswarship:
**Expected Behavior**
You can edit the metadata for an Individual and save your changes.

**Current Behavior**
If you edit the metadata for an individual and then click Save, nothing happens and the edit window stays up.

**Community Link**
https://community.wildme.org/t/cant-edit-metadata-on-individual-page-in-codex/2262

**QA Notes**
Weird error on submit. In testing with [this zebra](https://zebra.wildme.org/individuals/480fa440-e5cb-4a2e-a519-10ba87db7bc8), we couldn't save the sex or adoption name.

![Image](https://user-images.githubusercontent.com/103530016/255730199-64201367-0562-4c14-a2df-dbb3c9d5a852.png)

- Tested with [this seal](https://qa.dyn.wildme.io/individuals/d7a5dd31-c604-4ba2-961b-c50f5bd32700) in QA server; could not reproduce behavior.
- Tested with [this seal](https://seals.wildme.org/individuals/9d3b9514-20b0-486d-8982-a24edc9e3f34) in Seal Codex; could not reproduce behavior.

PR #541, opened 2023-08-17
hotfix of issue 523

erinz2020:
hotfix for #523 
#523 
by adding default value assignment, prevent program from throwing error when there is no custom fields in individual/sighting/encounter category.


naknomum (approved):
lgtm

PR #542, opened 2023-08-25
Autogen merge

erinz2020:
enable "codex ID" on merge page.
two codex ID will be populated, and user should choose one.
if one individual has codex ID but the other not, user has to choose the only one.
if both individual have no codex ID, unable to merge. this kind of data should not exist after enable codex ID.

when setting up codex ID on an exiting server, first step is to go to species management -> edit prefix page, choose one species to edit its prefix, and all the rest species will be set a temp prefix automatically.


naknomum (approved):
looks good

PR #543, opened 2023-08-25
add info box on create individual page

erinz2020:
add an info box on create individual page about the codex ID


naknomum (approved):
lgtm

PR #544, opened 2023-08-25
autogen-create-individual-page

erinz2020:
add an info box on create individual page about codex ID
adjust the "CONFIRM MATCH" button's border radius to get aligned with other buttons


naknomum (approved):
lgtm

Issue #534, opened 2023-07-31
change log: search


TanyaStere42:
By giving admins and support access to a table of the changes that have taken place in the platform, we are enabling them to see patterns and problems that may improve troubleshooting and head off development and support issues

## Resources

Final design: https://www.figma.com/file/9kvaU99j6Q5XjzSYTyWmtZ/Change-Log?node-id=124-448&t=3hFNSNDVGWvYqyPo-0

![Image](https://user-images.githubusercontent.com/62726866/257335824-6da981f9-243e-4cf4-bb24-a8ff7313ec11.png)

## Technical notes

- we are not making modifications to the log APIs as part of this feature. This feature is entirely about getting users access to the logs without having to leverage developer tools

## User Stories

- [ ]  As an administrator on the change log page, I can input a guid to see a paginated list of actions that have been taken on that object

## QA

- [ ]  search the change log using a guid for each of the following object types. make sure multiple messages appear:
    - [ ]  sighting
    - [ ]  encounter/animal
    - [ ]  individual
    - [ ]  user

TanyaStere42:
qa validated!
notes: we may want to adjust this to be an ES table at some point but for now, this works great. Need to make a note in the documentation that page reloads on search

Issue #533, opened 2023-07-31
change log: page and table


TanyaStere42:
By giving admins and support access to a table of the changes that have taken place in the platform, we are enabling them to see patterns and problems that may improve troubleshooting and head off development and support issues

## Resources
Final design: https://www.figma.com/file/9kvaU99j6Q5XjzSYTyWmtZ/Change-Log?node-id=124-448&t=3hFNSNDVGWvYqyPo-0

![Image](https://user-images.githubusercontent.com/62726866/257335608-9904e86f-874c-476d-a48f-e0462bcb0c23.png)

## Technical notes

- we are not making modifications to the log APIs as part of this feature. This feature is entirely about getting users access to the logs without having to leverage developer tools

## User Stories

- [x]  As an administrator on the change log page, I can see a paginated list of all actions that have taken place on the platform

## QA

- [x]  Verify the change log page is populated with a table that lists messages of what has happened in the platform
- [x]  change a user’s name. verify the change shows in the table
- [x]  create a sighting. verify it shows in the table
- [x]  perform matching. verify it shows in the table
- [x]  create an individual. verify it shows in the table
- [x]  create a collaboration. verify it shows in the table

TanyaStere42:
table display issue: can be worked around by copying text and pasting into an external text editor, so believe it's just a text-wrap issue


![Image](https://github.com/WildMeOrg/codex-frontend/assets/62726866/53b64c45-767d-4b80-8b46-311e047dedac)

erinz2020:
there is a scroll bar under the table, slide it to the right all the content can be displayed. but this is ugly, I'll make it wrap when overflow. going to push a patch later.

![1694114510305](https://github.com/WildMeOrg/codex-frontend/assets/99415659/e96bf799-2278-4e58-9ce7-f85c3a4ad14f)

TanyaStere42:
component issue; closing

Issue #532, opened 2023-07-31
change log: control panel


TanyaStere42:
Admins and support need to have access to changes in the platform so they can better support users

## Resources

Final design: https://www.figma.com/file/9kvaU99j6Q5XjzSYTyWmtZ/Change-Log?node-id=124-448&t=3hFNSNDVGWvYqyPo-0

![Image](https://user-images.githubusercontent.com/62726866/257334003-cb30a3c5-4354-48a8-84b5-571db869e48b.png)

Technical notes:

- we are not making modifications to the log APIs as part of this feature. This feature is entirely about getting users access to the logs without having to leverage developer tools

User Stories:

- [ ]  As an administrator on the control panel page, I can see a button that takes me to the change log page.
- [ ] URL: /settings/changelog

QA:

- [ ]  Verify the button is on the control panel and it navigates to an appropriate url

TanyaStere42:
QA verfied, used checklist above

PR #545, opened 2023-08-30
Changelog 1

erinz2020:
covers issue #532 #533 #534 

1. on control panel page, clicking at change log button will lead to /settings/changelog page
2. only admin users can see this page, else will show forbidden page
3. all change logs will be displayed and sorted in desc order by created date and time
4. user can search by item guid


naknomum (approved):
looks good

PR #548, opened 2023-09-05
fix the flashing setup page issue

erinz2020:
fix the flashing setup page issue. when data is still loading, show the loading screen, not the initiation page.


naknomum (approved):
lgtm

Issue #547, opened 2023-08-31
terms and conditions PDF not working on production


TanyaStere42:
## current behavior
user clicks TOU link, tab opens but 404 error because pdf cannot be found

## expected behavior
when user clicks TOU link, the pdf successfully opens

## investigation notes
problem stems from where the pdf is located on production environments vs dev environments

TanyaStere42:
QA Checklist
- [x] TOU displays on QA server. Alerting systems for verification on production server

PR #549, opened 2023-09-06
Fix TOU file not found issue on prod issue

erinz2020:
put the pdf file under /src/copy_on_build/, so when we are building the project, all the files under /src/copy_on_build/ will be copied to /dist directory which is the public directory of this project (reasonably inferred from the configuration file webpack.common.js) , so that all the static files under the output directory can be found. 

covers issue #547


naknomum (approved):
looks good

PR #551, opened 2023-09-11
Autogen name search

erinz2020:
1. implement searching for individuals by codex ID, case-insensitive, supports partial search
2. implement searching for sightings by codex ID, case-insensitive, supports partial search
3. fix a bug cause by case sensitivity in searching for individuals by first name/adoption name, now it's case-insensitive too.


naknomum (approved):
looks good

PR #557, opened 2023-09-15
Sightings individuals download

erinz2020:
enable search result download for individuals/sightings,
now get the excel file from backend, instead of making up the data from FE.
if no data to export, user will get an empty excel file.


naknomum (approved):
looks good

Issue #555, opened 2023-09-12
Sighting search - Encounter field UI integration


TanyaStere42:
As a researcher or admin on the sighting search page, I can search using any custom encounter field, or using the default encounter fields species and sex.

TanyaStere42:
Dependency on https://github.com/WildMeOrg/houston/issues/899 and https://github.com/WildMeOrg/houston/issues/900

TanyaStere42:
QA:

- [ ] displays in the ui
- [ ] in the export

PR #559, opened 2023-09-19
555-sighting-encounter-search

erinz2020:
1. now user can search sightings by encounter species/sex/custom fields.
2. abstract logic about checking the type/filter of custom fields to a hook and reuse the hook in sightings/encounters/individuals search schemas
3. fix a lint error about nesting ternary expressions
4. comment out some code about displaying encounter images count and individual info on sightings search page, because currently we don't have that in BE returned data

covers issue #555


naknomum (approved):
looks good

PR #569, opened 2023-10-02
562-563-last_region-and-last_freeform

erinz2020:
on individuals search result page, add last_region and last_freeform to individuals table right after last_seen


naknomum (approved):
looks good

PR #574, opened 2023-10-12
sightings metadata redesign

erinz2020:
sightings metadata page:
1. remove photograph tab
2. on annotations tab, display images with or without annotations
3. add "add annotations" button to the 3dot button
4. when click at an image that has no annotations, "add tag" is hidden, we can't add tag to image that has no annotation


naknomum (approved):
looks good

PR #575, opened 2023-10-16
536-match-page-set-state

erinz2020:
1. add dropdown on both sightings/match result page, user can select a status to update
2. remove the "mark sighting as reviewed" from the 3dot menu on sightings page
3. the dropdown has 4 options: reviewed, in progress, unreviewed, unidentifiable


naknomum (approved):
looks good

PR #576, opened 2023-10-18
503-nested-regions

erinz2020:
on sightings "re-run identification" page and and "commit sighting custom identification" page, when select a top level region, its sub regions will be all included.


naknomum (approved):
lgtm

PR #577, opened 2023-10-18
match-status-patch

erinz2020:
add 'unprocessed' status for uncommitted sightings on sightings page, the button will be disabled if the sighting is not committed.


naknomum (approved):
looks good

PR #578, opened 2023-10-19
restrict access right to bulk import

erinz2020:
only admin/staff/researcher can see the bulk import button on actions pane


naknomum (approved):
looks good

PR #579, opened 2023-10-24
535 no match confirmation

erinz2020:
1. add a "CONFIRM NO MATCH" button on sightings matching page
2. if selected annotation/animal has been assigned an individual, this button will be hidden
3. if selected annotation/animal hasn't been assigned individual, when click at the "CONFIRM NO MATCH" button, will create a new individual using this animal
4. "CONFIRM MATCH" button works as usual


naknomum (approved):
looks good

PR #592, opened 2023-10-25
increase confirm match button zIndex

erinz2020:
increase confirm match button zIndex so that it won't be blocked by table header


naknomum (approved):
easy!

PR #595, opened 2023-10-26
change photographs to annotations

erinz2020:
change photographs to annotations since photographs is not in use anymore


naknomum (approved):
good

PR #597, opened 2023-10-26
change add to assign to keep consistent with assign button

erinz2020:
on assign annotations to animal page dialog, change ADD to ASSIGN, to keep consistent with ASSIGN button


naknomum (approved):
looks good

PR #600, opened 2023-10-27
571-add-assign-button

erinz2020:
add an ASSIGN button in pending banner on sightings page, when user clicks at it, will navigate to animal tab, where user can assign annotations to animals.


naknomum (approved):
looks good

PR #601, opened 2023-11-01
516 unmounted component warning

erinz2020:
1. in useFetch hook, return statusCode directly from useQuery result, avoid state updates after a component has been unmounted.
2. remove some deprecated lines from control panel page
3. remove some unused imports from Requests page


naknomum (approved):
look good

PR #602, opened 2023-11-03
517 firefox performance issue

### LINKED ISSUE #517: performance issue blocks report sighting on firefox
**Expected Behavior**
User can submit a sighting on Firefox

**Current Behavior**
For some users, a series of 401 errors locks up the reporting page on firefox once you get past image submission stage. Suspected to be a performance issue

Image

erinz2020:
fix a performance issue on report a sighting page.
change the dependencies of useEffect from array to string, so that when the dependencies remain the same value, don't execute the setState inside useEffect.


naknomum (approved):
looks good

PR #612, opened 2023-11-14
588/589/590/591 animal search

erinz2020:
add animal button/animal search page/animal default fields/custom fields search, similar to sightings/individuals page


naknomum (approved):
looks good!

PR #615, opened 2023-11-29
Region editor remodel

erinz2020:
1. remodel region editor, UI looks different
2. now when user reports sighting and select region, the dropdown list is a tree view, we can see all regions and sub regions in tree structure
3. when selecting top level region, its sub regions will be expanded
4. when searching region, all regions includes the search text will be displayed, user can also expand the top region to see its sub regions, even if the sub regions don't match search text
5. this component is also used on "re-run identification" page and "customize identification" page
6. on the two pages above, it is still not multiple selection, but we will run job with the selected region and its sub regions, a previous ticket implemented this feature, not affected by this ticket.


naknomum (approved):
looks good

PR #621, opened 2023-12-18
610 image carousel

erinz2020:
a patch of image carousel: 
1. all images of the same individual/encounter will be displayed, no matter the image is in the query table/candidate table or not.
2. apply lazy loading of images: only images in the viewport will be loaded.


naknomum (approved):
looks good

PR #622, opened 2023-12-18
609 user data page

erinz2020:
a new page of "all my data":
when user select "my sightings", "my pending sightings" and "my sightings" will display
when user select "my unapproved sightings", "my pending sightings" and "my unapproved sightings" will display


src/pages/individual/SearchIndividuals.jsx
```diff
@@ -19,6 +19,8 @@ export default function SearchIndividuals() {
     reverse: true,
   });
 
+  console.log('SearchIndividuals.jsx: formFilters: ', formFilters);
```
naknomum, line 22: _might_ be left-behind debugging. but i am fine with leaving this if there is some on-going need to keep an eye on this.

naknomum (approved):
looks good to me. one tiny comment about a `console.log()` in there; but no biggie.

src/pages/individual/SearchIndividuals.jsx
erinz2020: thank you. removed.

PR #624, opened 2023-12-27
620 region administration

erinz2020:
region administration:
1. change the region editor from a modal to a page
2. update layout, add finish and cancel button
3. add a checkbox in front of every region: user can select and deselect any region and save
4. when creating sighting/re-run identification, deselected regions will be disabled and can't be chosen


naknomum (approved):
looks good to me

PR #625, opened 2023-12-28
data page paginating

erinz2020:
add paginating to my sightings/my pending sightings/my unapproved sightings


naknomum (approved):
looks good

PR #627, opened 2024-01-04
region administration

erinz2020:
region administration:

- change the region editor from a modal to a page
- update layout, add finish and cancel button
- add a checkbox in front of every region: user can select and deselect any region and save
- when creating sighting/re-run identification, deselected regions will be disabled and can't be chosen


naknomum (approved):
looks good

PR #629, opened 2024-01-04
623 export restricted access

erinz2020:
deal with response status code 400 and 403 when making requests to download search result on individuals/sightings/encounters page.
if status code is 400 or 403, a dialog will open with messages indicating there is no result to export etc.


naknomum (approved):
looks good

PR #632, opened 2024-01-09
remove exporter from UI and optimize code

erinz2020:
remove exporter role from UI:
1. user won't be able to create exporter role
2. also remove exporter role on edit users page
3. update the criteria to hide/display download button: now everyone can see the button, any user can possibly download, but with different access right to data.


naknomum (approved):
looks good

PR #635, opened 2024-01-10
626 animal search page action directing

erinz2020:
1. on animal search page, user can click to navigate to the sighting: animal tab of a given animal.


naknomum (approved):
seems good

PR #639, opened 2024-01-25
604 export collaboration

erinz2020:
add export level access for users collaboration, and restrict edit of user profile
1. user can request/approve/deny/revoke export collaboration
2. manager can set export collaboration between two users
3. notification also reflects export level requests
4. view access blocks export access, export access blocks edit access
5. edit user metadata button can only be seen by user himself and manager
6. edit preferences button can only be seen by user himself


naknomum (approved):
good by me

